PR #7, opened 2019-11-03
Add src/runtime/devmode.r

Don-Ward:
devmode.r is a file to store code used during development that is not
intended to be available if the system is configured  without the
--enable-devmode option.

### DIFF
--- a/src/h/fdefs.h
+++ b/src/h/fdefs.h
@@ -31,6 +31,9 @@ FncDefV(constructor)
 FncDef(copy,1)
 FncDef(cos,1)
 FncDef(cset,1)
+#ifdef DEVMODE
+FncDef(dbgbrk,0)
+#endif /* DEVMODE */
 FncDef(delay,1)
 FncDefV(delete)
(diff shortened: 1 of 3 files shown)


src/h/fdefs.h
```diff
@@ -31,6 +31,9 @@ FncDefV(constructor)
 FncDef(copy,1)
 FncDef(cos,1)
 FncDef(cset,1)
+#ifdef DEVMODE
+FncDef(dbgbrk,0)
+#endif /* DEVMODE */
```
Jafaral: Not a big deal, but the common format for the closing comment is to leave 5 tabs after endif.

Don-Ward:
OK, I’ll fix it as an exercise. But tabs? really? I’ve been doing  M-x untabify on most of my new stuff to avoid the “displaying differently” problem.

Jafaral:
@Don-Ward I use "untabify" as well for the same reason you mentioned, but I kept this one to keep it consistent with all other places in the code.

PR #12, opened 2019-11-08
Enable larger strings on 32-bit and 64-bit systems

Jafaral:


Jafaral:
@Don-Ward It is a tradition, slightly more readable than hex I think?

PR #21, opened 2020-01-08
Fix a problem with using space-separated IPATH and LPATH settings

Don-Ward:
Patch supplied by Steve Wampler; see
  https://sourceforge.net/p/unicon/mailman/unicon-ldif/?viewmonth=201912


cjeffery:
Looks fine to me. Officially, replace() should be map(). If one was to stick with replace(), what in this file guarantees that the strings module will get linked in? I missed that.

PR #46, opened 2020-04-20
Fix GC deadlock in initial expressions.

Don-Ward:
If a thread that is in an initial expression causes a garbage collection
whilst another thread is waiting to execute an initial expression then a
deadlock occurs (because the waiting thread never enters the GC queue,
so the GC thread hangs waiting for all threads to be stopped and
NARthreads to drop to 0).

### DIFF
--- a/src/runtime/interp.r
+++ b/src/runtime/interp.r
@@ -2054,7 +2054,7 @@ L_agoto:
 
 	 case Op_Init:		/* initial */
 #ifdef Concurrent
-	    MUTEX_LOCKID_ALWAYS(MTX_INITIAL);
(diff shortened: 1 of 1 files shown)


Jafaral:
The locking mechanism evolved over time to optimize for non concurrent programs as well as reduce the overhead when possible. Short/fast critical regions with no possibility of GC's or other locks inside the critical region can be safely executed with a simple lock/unlock without touching NARthreads. It just means all threads wanting to enter that  critical region would all have to enter/exit the critical region before a GC can take place on the assumption that the GC case would be the less common one. Thank you @Don-Ward for this fix. @cjeffery we can definitely review other places to make sure we always do the appropriate type of lock in every place. I mus have overlooked this one for this initial block as it was added late in the game.

PR #48, opened 2020-05-05
1st draft of the Heap Verification routines.

Don-Ward:
All verification is controlled by an environment variable VRFY.
Some structures, notably lists, are not verified yet.
See rmemmgt.r for details.

### DIFF
--- a/src/h/rexterns.h
+++ b/src/h/rexterns.h
@@ -419,3 +419,9 @@ extern struct  b_cons *LstTmpFiles;
 #ifdef Audio
 extern int isPlaying;
 #endif					/* Audio */
+
+#ifdef VerifyHeap
+extern long vrfy;
+extern void vrfyLog(const char *fmt, ...);
+extern void vrfy_Live_Table(struct b_table *b);
+#endif                  /* VerifyHeap */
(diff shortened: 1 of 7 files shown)


src/h/rexterns.h
```diff
@@ -419,3 +419,9 @@ extern struct  b_cons *LstTmpFiles;
 #ifdef Audio
 extern int isPlaying;
 #endif					/* Audio */
+
+#ifdef VerifyHeap
+extern long vrfy;
+extern void vrfyLog(const char *fmt, ...);
+extern void vrfy_Live_Table(struct b_table *b);
```
Jafaral, line 426: Where are these functions? I don't see their implementation as part of this PR?

Jafaral: Found it! Github wasn't showing rmemmgt.r by default because  the diff is large!

Jafaral:
@Don-Ward if you want the code under your new VerifyHeap to be CI tested on github with out messing with the CI test itself , you can do the trick of  defining VerifyHeap as part of the the commit and push it to a temporary branch on your sandbox.

Jafaral:
Just did that here: https://github.com/Jafaral/unicon/commit/59775d1f4d538059a4eb25543ff567dfec056209

Jafaral:
Test failed on Ubuntu: https://github.com/Jafaral/unicon/runs/650353668?check_suite_focus=true

Don-Ward:
The error happens on the Ubuntu build because the verifier code references something that is not defined when we are building iconc (i.e. when COMPILER is defined). I removed the offending code in commit 7eb3098.

A more interesting question is why the macOS build does not build iconc.

Jafaral:
For bookkeeping it is probably good to say why we are closing this one and not merging, something like:

Merged with #49

PR #56, opened 2020-05-15
Adjust Unicon revision to correspond to SVN's revision (take 2).

Don-Ward:
The calculation of the SVN revision number (stored in src/h/revision.h)
was being overridden by the update_rev target in src/runtime/Makefile.
That target is now active only if the file src/h/revision.h does not exist.
A missing newline has been added to src/h/revision.h. and the file is now
removed by the Pure target (since it's a configured file).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -345,13 +345,13 @@ update_rev:
 	   echo "#include <stdio.h>" > plus4.c; \
 	   echo "int main(void)" >> plus4.c; \
 	   echo "{" >> plus4.c; \
-	   echo "   printf(\"#define REPO_REVISION \\\"%d-%s\\\"\"," \
+	   echo "   printf(\"#define REPO_REVISION \\\"%d-%s\\\"\\\n\"," \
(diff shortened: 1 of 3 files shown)


src/Makefile
```diff
@@ -73,7 +73,7 @@ clean Clean:
 
 
 Pure: Clean
-	$(RM) h/auto.h
+	$(RM) h/auto.h h/revision.h
```
Jafaral, line 76: This will break out-of-tree builds. The only reason revision.h was introduced is to preserve revision info for when you create a source distribution (`make dist`) where you don't have git data anymore

Jafaral: i.e, you don't want someone building from source distribution ending up with no revision.h accidentally when the do `make clean`

Don-Ward: OK, I'll take it out.

Jafaral:
@Don-Ward can you please just squash the new commit back into the original commit and force push ?

PR #64, opened 2020-05-26
Revert "Adjust Unicon revision to correspond to what SVN's revision w…

Jafaral:
…ould have been."

This reverts commit 7295925164b18d83d6a4f0e0ba3c01dd7f710e3a.

There is no need to align with svn revision numbers. All commits are within git context now.
The revision number should reflect git's history, not svn history.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@
 #  configuration parameters
 VERSION=13.1
 name=unspecified
-REPO_REV_COUNT="$(shell LC_ALL=C git rev-list --count HEAD )"
+
+REPO_REV_COUNT="$(shell LC_ALL=C git rev-list --first-parent --count HEAD )"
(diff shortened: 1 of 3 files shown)


Jafaral:
@Don-Ward I used to be that there is only one place to calculate revision info: top level makefile, that is where most user  use to build Unicon. For developers/users who might "live" inside src/runtime, they need to udpate rev info if they build from that location.  So, the recipe is needed in both places. In theory we can refactor  some of that into top-level Makedefs, but that is not read by Makefile at the moment because the first time you run make at the top level you wont have Makedefs yet because it is auto-generated by the configuration script. We could make it a requirement to run the configure script before  you can do your first "make", but Unicon users were accustomed
 to running make out of the box and I left it as-is, for now :)

Jafaral:
BTW, an addition to syncing the two versions to be the same, I also added the option to git to calculate revision info using first parent.

PR #69, opened 2020-06-18
Replace the terms master/slave with main/secondary

Jafaral:


brucerennie:
As Don has said, primary is the preferred term. Reading the code changes, it is actually more comprehensible to have all changes to "main" changed to "primary". This provides a clearer semantic handle on what the code is actually representing.

Not to be offensive, if this change is about the current political correctness then it is not worth the effort. Master/slave or primary/secondary give the same technical designation as far as what is happening in the code.

PR #72, opened 2020-07-07
Uni lib

StephenWampler:
Two changes to uni/lib:
   -- complex.icn  -- fixed typos in comments ("scaler" -> "scalar") as caught by Don
   -- predicat.icn -- extend istype so 2nd argument can be a class instance as well as a class name.  This affects
               instanceof() [same file] and instanceOf() [object.icn] to test against the class instance, as in:

            if X.instanceOf(Object()) then ...

      complementing the existing equivalent:

           if X.instanceOf("lang::Object")

### DIFF
--- a/doc/ib/appH.tex
+++ b/doc/ib/appH.tex
@@ -534,85 +534,100 @@ ZPIPE & 9 & uncompress .Z files via a piped process (libXpm) \\
 
 \newpage
 \section{A Bestiary of Unicon Releases}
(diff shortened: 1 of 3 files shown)


Jafaral:
@StephenWampler  you need to start with a clean branch checkout out of the upstream master branch, then add the last two commits only. The  issue you had before and this one is due starting with one of your local branches rather than the master branch from the upstream project.

PR #77, opened 2020-07-11
Add some missing files to Makefile.

StephenWampler:
Teaches make about a number of .icn files in uni/lib.

### DIFF
--- a/uni/lib/Makefile
+++ b/uni/lib/Makefile
@@ -4,7 +4,7 @@ include $(BASE)/Makedefs.uni
 .PHONY: all clean deps
 
 UFILES=gui.u file_dlg.u db.u \
- address.u args.u base64handler.u basicclasscoding.u blockread.u class.u classcoding.u \
+ address.u base64handler.u basicclasscoding.u class.u classcoding.u \
  cltable.u comparator.u complex.u compoundedit.u contentdisposition.u contenttype.u \
  cvsuser.u cvsutil.u decode.u encode.u encodinghandler.u error.u exception.u \
  fcn_util.u format.u \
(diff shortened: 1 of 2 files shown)


uni/lib/Makefile
```diff
@@ -4,13 +4,14 @@ include $(BASE)/Makedefs.uni
 .PHONY: all clean deps
 
 UFILES=gui.u file_dlg.u db.u \
- address.u base64handler.u basicclasscoding.u class.u classcoding.u \
+ address.u args.u base64handler.u basicclasscoding.u blockread.u class.u classcoding.u \
```
Jafaral, line 7: args.u imports lang which has to be built first. Looks like runnable.u is the last file to be built which is part of lang so I suggest moving arg.u after it

uni/lib/Makefile
```diff
@@ -4,13 +4,14 @@ include $(BASE)/Makedefs.uni
...
- cvsuser.u cvsutil.u decode.u encode.u encodinghandler.u error.u format.u \
+ cvsuser.u cvsutil.u decode.u encode.u encodinghandler.u error.u exception.u \
+ fcn_util.u format.u \
  group.u httpclient.u json.u langprocs.u listener.u \
- mailbox.u netclient.u mailmisc.u message.u messagehandler.u method.u \
- money.u msg.u multipart.u multiparthandler.u noophandler.u object.u \
- popclient.u predicat.u process.u pushback.u qsort.u quotedprintablehandler.u \
+ mailbox.u mailmisc.u mapbytes.u message.u messagehandler.u method.u \
+ money.u msg.u multipart.u multiparthandler.u netclient.u noophandler.u object.u \
+ popclient.u predicat.u process.u property.u pushback.u qsort.u quotedprintablehandler.u \
```
Jafaral, line 14: Not sure if any of the remaining files would have the same or similar dependency issue.  I suggest trying to build locally first and fix any issue if there are any.

StephenWampler:
Replaced with corrected pull request.

PR #81, opened 2020-07-14
Separate Notifier API from Connectable API - behavior is different.

StephenWampler:
Pulled the code taken from UniLib into a separate notifier.icn class as the behavior differs from that in connecctable.icn (see comment at start of notifier.icn).  Updated Makefile to build notifier.u.


uni/lib/str_util.icn
```diff
@@ -14,6 +14,8 @@
 
 package util
 
+import lang
+
```
Jafaral, line 18: Why is this import needed here?

Is it OK for packages util and lang to import each other?

StephenWampler: Because str_util contains calls to makeProc{}, which is defined in lang.  I don't see why two packages can't
import each other - so long as there's no cycle in the actual code execution.  Maybe there's a reason, but I
have sample programs that won't work without this change and work fine with it.

Jafaral:
@StephenWampler, I see you keep pushing new commit to this PR and making it better. Let me know when you think it is in a good shape and ready to merge.

StephenWampler:
On 7/19/20 9:12 PM, Jafar Al-Gharaibeh wrote:
>
> @StephenWampler <https://github.com/StephenWampler>, I see you keep 
> pushing new commit to this PR and making it better. Let me know when 
> you think it is in a good shape and ready to merge.
>

"making it better" - what a polite way of saying "fixing your errors"!  
I'm done (he hopes).  I've got all of my demonstration programs that 
used to work
with the old UniLib now running using the uni/lib equivalents - except 
for two that use GUI features that I need to figure out what's 
different.  Neither
of those demos, however are something I'd put into uni/* however.

Thanks,
Steve

-- 
Steve Wampler - sbw@tapestry.tucson.az.us
The gods that smiled at your birth are now laughing out loud - fortune cookie

uni/lib/str_util.icn
cjeffery: I think there is a chicken-and-egg bootstrapping issue with packages importing each other. If both import each other, whichever one you compile first will fail because the other package doesn't exist yet. In practice, one can get it bootstrapped by starting with partial or near-empty packages, but it will require multiple compiles to get all the symbols into the packages in the .gdbm files, and then recompiling, before the imports will be fully correct. Kind of a Mr. Yuk, but viable as long as the makefiles are smart enough to bootstrap.

PR #82, opened 2020-07-16
Don't overlap flags with large string size

Jafaral:

### DIFF
--- a/src/h/cpuconf.h
+++ b/src/h/cpuconf.h
@@ -13,7 +13,7 @@
  */
 
 /*
- * Most of the present implementations use 32-bit "words".  Note that
+ * Most of the present implementations use 64-bit "words".  Note that
  *  WordBits is the number of bits in an Icon integer, not necessarily
  *  the number of bits in an int (given by IntBits).  For example,
  *  in MS-DOS an Icon integer is a long, not an int.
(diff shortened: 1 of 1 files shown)


Jafaral:
@Don-Ward aside from Raspberry Pis, 32-bit are almost extinct, and every those are moving to 64-bit anyway. So for most practical situations, we are fine and safe with the limit on 64-bit platforms. 

Just like you, when I  increased the limits a few months ago after reviewing the spec's/design in Icon't and Unicon implementation book. I did write some test application and it worked. I tried to do the same for recently for Icon. but Gregg comment was how does that affect the flags and pointed at a specific line in the interpreter code which looked fine to me at the face of it. Then I looked at macros that test the flags like:

https://github.com/uniconproject/unicon/blob/master/src/h/rmacros.h#L294

These are not making any test to make sure the descriptor is not a "qual" before proceeding to test the flags like I assumed given the documentation. We could go and fix that, but given what I said about 32-bit system being almost extinct anyway, I'd rather just leave them alone and not risk changing some assumption in the existing code. Stings that large are rare also anyway. 

> Is this change covering up a bug somewhere else (i.e. code that doesn't check the n flag is set before checking the value of other flags)?
   a bug or a feature as you always say? after all, the code runs a bit faster when you have to make one test instead of two to determine that type of a descriptor. ;-)

Don-Ward:
I think I agree with the conclusion on the related Icon PR; i.e. leave the 32bit code as is and (if you have to) make the change for 64 bit platforms only. I still don't think it should be necessary.

PR #89, opened 2020-07-25
Revert Args.icn to UniLib behavior

StephenWampler:
This reverts the behavior of Args.icn to the original UniLib behavior.  The indent is to separate options (command line parameters that start with one or more hyphens) from arguments (parameters that don't), simplifying the process of handling options as well as arguments that aren't options (e.g. lists of file names).


Jafaral:
@StephenWampler, I see two merge commits on this PR that aren't supposed to be here. Can you fix that please ?

StephenWampler:
Do you mean efd9342 and bd1017d?  I don't know how they got there into this PR!  Is this because master got ahead of ulib before the commit?  Anyway, I'm going to close this PR and try opening another...

PR #93, opened 2020-08-04
Add to Makefile.in two lines to clear IPATH and LPATH to avoid

brucerennie:
conflicts with user defined IPATH and LPATH environment variables.


Jafaral:
@brucerennie, can  you add the IPATH to LPATH to Makefile as well ? I like to keep Makefile and Makefile.in in sync. 

Thanks!

PR #96, opened 2020-08-05
Correct spelling mistake in overload message (Overlolading to Overloa…

brucerennie:
…ding).

### DIFF
--- a/configure
+++ b/configure
@@ -15608,7 +15608,7 @@ else
 fi
 
 if test "x$ovld"      = "xyes" ; then
-  echo " + Overlodading"
+  echo " + Overloading"
 else
   echo " ! Overloading"
 fi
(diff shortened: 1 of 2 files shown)


Jafaral:
Hi @brucerennie , did you change `configure` by hand or did you run `autoreconf -i` to generate it ?   I ask because I never change it by hand, I always  make changes to `configure.ac` and then run `autoreconf -i` to generate `configure` and then commit both at the same time.

brucerennie:
Good morning Jafar. I changed both files by hand, starting with configure.ac and then configure. So, other than the automatic removal of spaces and tabs from the end of various lines in configure.ac by my editor, the only significant change was the removal of the extra character from the same message in each file. 

If you would like me to run the autoreconf -i command and then resubmit this, I will do so. I would however suggest that some form of documentation is placed in the README.md files to indicate that this needs to be done for those of us who do not use these tools as a reminder of what we need to do to make sure that the normal project rules are followed.

Jafaral:
It doesn't matter if you change it by hand as long as it matches what the standard autoconf tool generates. I  just checked, it matches, I will be merging this shortly.

PR #103, opened 2020-08-11
Correcting Makefile install "docs and mans"

brucerennie:
The Makefile references README in the Install section instead of README.md.

This fails to copy the README.md document to the documentation directory
when installing.


brucerennie:
The force push was required to fix a typo from "," to "." in the Makefile for README.md.

PR #105, opened 2020-08-13
Updates for the Test Suite

brucerennie:
When running the test suite, a number of programs failed in the general
and the posix test areas. In the case of the general test area, it was
found that one of the IPL procedures needed to be altered. This has
affected the file numbers.icn.

Previously, a new version of large() was added to handle the versions
of Unicon that had the concurrency facilities added. Unlike the old
version of large(), which is still retained in numbers.icn, the new
version failed to return the supplied parameter. This change is now
implemented.

One of the tests in errors.icn, required a change of parameter value
to be used as the original no longer fails. The errors.std file was
updated to reflect that some of the error messages related to set
construction have changed to include tables.

For ilib.icn, the test value for large() was changed to use a value
that exceeds 128 bits in length. This is to cover any future implementation
changes that would use 128 bit words in the underlying machine. As a
consequence, the appropriate change was also made to ilib.std.

For the posix test system.icn, a change was made to the system call
in relation to the command line file redirection, as the original
use was not creating the correct output.

For the posix test pty_uni-icn, due to a memory violation runtime
abort, the flush() call was commented out. This has allowed the
program to run to completion. In addition, due to a newer version
of udb (from 2.0 to 2.1) the messages in pty_uni.std were also updated
to reflect this change.

The standard test output file has a slight difference to what is actually
written by the thread test mutexfun.inc. In the standard test output the
value of x is given as 46 whereas the actual output is 45.

If you add into the source code a copy of the line

x := mutex()

then the output matches.

The standard test output file has been updated to reflect this change in
value to 45 instead of 46.


tests/posix/system.icn
```diff
@@ -44,5 +49,10 @@ procedure main()
 #      }
    remove("foo.out")
    remove("x.icn")
+   #
+   # add following remove command to remove the generated file, caused
+   # problem in Makefile
+   #
+   remove("x")
```
Jafaral, line 56: will this work on Windows, since the generated executable would be  "x.exe" ?

I didn't check to see how much the posix test has coverage on Windows anyway.

tests/posix/stand/pty_uni.std
```diff
@@ -17,102 +17,98 @@ pty gives me: "hello, world"
 "   UDB Version "
 "   UDB Version 2"
 "   UDB Version 2."
-"   UDB Version 2.0"
```
Jafaral, line 20: We should find away to ignore this part to avoid having to update the test output every time UDB version changes, although that rarely happens.

tests/posix/pty_uni.icn
```diff
@@ -32,26 +32,35 @@ $endif
...
 	   write(image(line))
+        }
        if find("(udb)", line) then break
    }
    write("got initial udb prompt")
-   write(fi, "where"); flush(fi)
+   #
+   # remove flush() call as it is causing runtime memory violation
+   #
+   write(fi, "where"); #flush(fi)
```
Jafaral, line 50: Maybe we should fix flush() instead of hiding it under the rug

tests/posix/system.icn
Jafaral: Something like this would do:
```
$ifdef _MS_WINDOWS_NT
   remove("x.exe")
$else
   remove("x")
$endif
```

tests/posix/pty_uni.icn
Jafaral: I wonder if  this is a problem with using flush with pipes!

tests/posix/system.icn
brucerennie: I have inserted your suggested change here and will push this shortly.

tests/posix/pty_uni.icn
brucerennie: My c programming skills are not sufficient to investigate this matter at the moment. There is another matter with the pipes that I observed and that related to what the unicon program is seeing coming from the other end. When you senf a command to UDB, as you read the response, it appears to resend the same information many times with a terminating CR that would simply overwrite your display. Is this a feature of the pipes implementation or a bug?

An example in the data  transcript is the sequence:

" w\r         \r(udb) wh"
" w\r         \r(udb) wh\r"
The "\r" in effect returns the cursor to the same position each time and the following characters will overwrite what was there previously.   

When you run the UDB program, there does not appear to be this occurring. I think this needs investigation.

brucerennie:
In regards to the various comments you have already made, do you want me to split each of these commits into separate requests? I am happy to do that.

PR #109, opened 2020-08-22
Move exception.icn down in Makefile

StephenWampler:
exception.u depends on misc_util.u, build it afterwards.


Jafaral:
@StephenWampler  many of the commits overwrite  pervious commits  in the same PR. I suggest combining/amending the commits when you have work in progress fixing things as you go. Fewer commits in this case makes it easier to follow the change and the end results is less history to commit to git.

PR #110, opened 2020-08-25
Allow strings as function names

StephenWampler:
This allows the function utilities to invoke procedures, functions, and operators using their string names.


Jafaral:
@StephenWampler  the second commit undoes the  first  commit, can you please  squash that two commits into one and force push ?

PR #116, opened 2020-09-03
Creation of xoptions.icn - replacement for options.icn

brucerennie:
This file contains an extended (highly modified) version of a command
line argument processing procedure. It includes more POSIX and GNU
command line processing as well as additional changes and extensions.

For testing purposes, there are two files that are included in the
data directory. Uncomment the procedure main() in the xoptions.icn
file, compile and run. The output should be 44 lines of output.

### DIFF
--- a/ipl/procs/xoptions.icn
+++ b/ipl/procs/xoptions.icn
@@ -1,0 +1,861 @@
+############################################################################
+#
+#       File:     xoptions.icn
+#
+#       Subject:  Procedure to get command-line options
+#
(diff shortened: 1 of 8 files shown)


Jafaral:
@brucerennie Thank you for this excellent work.  Can you please split the test and make it standalone where we can add it somewhere under the tests directory following the format we have?  Not sure what is  the best place for it, maybe under tests/unicon, tests/samples? part of where it should go depends on whether the test is meant to validate the new function or demonstrate how it is used.

PR #124, opened 2020-09-17
Supply the required include definitions to appease clang v12.

Don-Ward:
On macOS the function sysctlbyname() is used in rmemmgt.r.
Clang v11 warned of an implicit definition (because the
needed includes were not present). Clang v12 raises an error.

### DIFF
--- a/src/h/posix.h
+++ b/src/h/posix.h
@@ -83,7 +83,7 @@ extern char *sys_errlist[];
 #include <sys/param.h>
 #endif
 
-#if (defined(BSD) || defined(BSD_4_4_LITE)) && !defined(MacOSX)
+#if (defined(BSD) || defined(BSD_4_4_LITE)) && !(defined(MacOSX) || defined(MacOS))
 #define Setpgrp() setpgrp(0, 0)
 #else
 #define Setpgrp() setpgrp()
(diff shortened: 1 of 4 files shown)


cjeffery (changes requested):
This should be addressed in src/h/sys.h where there is currently an #ifdef MacOSX that pulls in sys/sysctl.h. Perhaps someone nuked MacOSX without changing all the references to MacOS, or something like that.

src/runtime/rmemmgt.r
```diff
@@ -8,6 +8,8 @@
 #include <malloc.h>
 #endif					/* CRAY */
 
+
```
Jafaral, line 11: Looks like  you left some dangling white space here!

src/h/posix.h
```diff
@@ -83,7 +83,7 @@ extern char *sys_errlist[];
 #include <sys/param.h>
 #endif
 
-#if (defined(BSD) || defined(BSD_4_4_LITE)) && !defined(MacOSX)
+#if (defined(BSD) || defined(BSD_4_4_LITE)) && !(defined(MacOSX) || defined(MacOS))
```
Jafaral, line 86: Let us just drop MacOSX @Don-Ward . There is no reason to keep it around.

Jafaral (changes requested):
Just minor comments, Thanks Don!

Jafaral:
Addressed in #126

PR #129, opened 2020-09-22
Updated UniDoc for use with uni/lib

StephenWampler:
The UniDoc sources now compile and run using the uni/lib support instead of the old UniLib librareis. 
 Note that the build*.sh scripts require customization for each system they are run on to propertly locate the Unicon source hierarchy.

### DIFF
--- a/uni/unidoc/Makefile
+++ b/uni/unidoc/Makefile
@@ -1,4 +1,6 @@
 SHELL=/bin/sh
+BASE=../..
+include $(BASE)/Makedefs.uni
 
 ########################################################
 # Manage the construction of the UniDoc program        #
@@ -26,14 +28,14 @@ help:
 #    commonly used suffixes.
 #
 %.u:	%.icn
-	unicon -c $?
+	$(UNICON) $(UFLAGS) -c $?
 
 UniDoc: $(UFILES) UniDoc.icn
-	unicon UniDoc
+	$(UNICON) UniDoc
 	@date >.lastbuild        # remembers when things were last built.
 
 ufiles: $(SRCS)
-	unicon -c $?
+	$(UNICON) $(UFLAGS) -c $?
 
 doc docs:	UniDoc
 	./buildDocs.sh
--- a/uni/unidoc/Makefile
+++ b/uni/unidoc/Makefile
@@ -1,5 +1,4 @@
-BASE=../..
-include $(BASE)/Makedefs.uni
+SHELL=/bin/sh
 
 ########################################################
 # Manage the construction of the UniDoc program        #
@@ -9,9 +8,6 @@ include $(BASE)/Makedefs.uni
 #
 SRCS   := $(wildcard ./*.icn)
 UFILES := $(filter-out ./UniDoc.u,$(patsubst %.icn, %.u, $(SRCS)))
-CLIB   := ../../Classes
-
-all: ufiles UniDoc
 
 # The first target is the default target.
 #    The "@"'s tell make to suppress printing the command before
@@ -30,15 +26,14 @@ help:
 #    commonly used suffixes.
 #
 %.u:	%.icn
-	$(UNICON) $(UFLAGS) -c $?
+	unicon -c $?
 
 UniDoc: $(UFILES) UniDoc.icn
-	$(UNICON) $(UFLAGS) UniDoc
-	$(CP) UniDoc$(EXE) ../../bin/
+	unicon UniDoc
 	@date >.lastbuild        # remembers when things were last built.
 
 ufiles: $(SRCS)
-	$(UNICON) $(UFLAGS) -c $?
+	unicon -c $?
 
 doc docs:	UniDoc
 	./buildDocs.sh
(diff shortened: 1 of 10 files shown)


uni/unidoc/buildIPLdoc.sh
```diff
@@ -2,17 +2,19 @@
 
 # Build and install the IPL documentation
 
-SBASE=/opt/unicon/ipl
-TBASE=/var/www/html/unicon/ipl
+# You must reset UBASE for your system's location of Unicon.
+UBASE=/opt/unicon
```
Jafaral, line 6: Steve,  the configure scripts already gives you this location. See uni/ide/Makefile as an example: 

```
BASE=../..
include $(BASE)/Makedefs.uni
```

Makedefs.uni should provide any of such system dependent variables. If you need something that isn't there please let me know and I can add it. The end user shouldn't need to edit files individually.

Jafaral:
There is a huge white space diff in a few files, was the intentional ?

uni/unidoc/buildIPLdoc.sh
StephenWampler: Thanks.  Makefile.uni,however, isn't sourceable by a script, just Makefiles.  I've been reluctant to force the use of
the scripts into the make process because running both adds about 24MB of html files into the doc hierarchy.  If
folks are okay with that I can merge there calls into the Makefile.  However, simply setting UBASE to a relative path
will also work in the build scripts.  I'll do that for now.

StephenWampler:
> There is a huge white space diff in a few files, was the intentional ?

Interesting.  I build this new version from my own copies of UniDoc, not from the sources in the unidoc directory
and copied those overtop of the unidoc versions once I had things running.

So the differences may be a difference in the editors used to build the files.  I tink I'll try and get away with
claiming it was intentional...

Jafaral:
white space could be space vs tab issue. I didn't check. It is good to keep those in one form or the other but minimize flipping between the two. No need to register undesired changes in the commits over time.

uni/unidoc/buildIPLdoc.sh
Jafaral: We do have:
docdir=${datarootdir}/doc/${PACKAGE_TARNAME}
mandir=${datarootdir}/man
datarootdir = ${prefix}/share

But we can add one from html generated  docs. We can add a configure option to generate html docs with a configurable target install location for such files. Would that be a good option ?

StephenWampler: I like that.  One thing I'm trying to decide is how much source to process in the html generation.  Right now it's
just some of the directories in IPL (19MB of html) and uni/lib (5MB).  It's tempting to expand it further but that's
a lot of space...

Jafaral:
Steve, can you do a  rebase instead of merge when you update your PR branch ? This is the preferred method especially when  you have  a few  commits only in the PR. Doing a merge adds an unnecessary merge commit in this case and it just complicate the  commit history.

Jafaral:
BTW, I didn't forget about an html config option. Will add that this weekend.

StephenWampler:
> BTW, I didn't forget about an html config option. Will add that this weekend.

No rush - I've pulled this back into Draft to add a few features and fix a few annoyances...

StephenWampler:
Jafar,

I did a rebase of (local) unidoc onto origin/unidoc first, but that didn't seem to update origin/unidoc until
after I followed up with a merge of unidoc onto origin/unidoc (a couple of web pages give this sequence
as how to do a rebase).  Let me know if there was a better process!

Thanks.

uni/unidoc/Makefile
```diff
@@ -30,15 +26,14 @@ help:
 #    commonly used suffixes.
 #
 %.u:	%.icn
-	$(UNICON) $(UFLAGS) -c $?
+	unicon -c $?
```
Jafaral, line 29: Any reason to use Unicon instead if the variable being replaced ?

StephenWampler: I didn't see where $(UNICON) was defined anywhere.  Perhaps it's defined if this Makefile is invoked from the main
Unicon make (i.e. via ${UNICON}/make), but a direct make in the uni/unidoc directory fails.

Jafaral: All unicon/uni directories get their definitions from the top level Makefile.uni 

BASE=../..
include $(BASE)/Makedefs.uni

StephenWampler: Ah - the Makefile in uni/unidoc is missing the lines:

BASE=../..
include $(BASE)/Makedefs.uni

lines at the top.

uni/unidoc/buildIPLdoc.sh
StephenWampler: Would you like me to update the Makefile with the missing lines and switch unicon to $(UNICON)?

Jafaral: That would be better probably. This way it aligns with  how the rest of the libraries are built.

PR #130, opened 2020-09-24
Add a discussion of large integers to appendix A of the Unicon book.

Don-Ward:


cjeffery (approved):
While there are minor wordsmithings that could improve these changes, the main question I am left with is: how many of the C-integer-only limitations are just artifacts of the current implementation, and how many are inherent and never likely to be fixed. Because when we document them like this, it makes it sound like they are language canon rather than artifacts of the current implementation.  Example: it is not obvious that i to j by k or seq() will not get extended whenever someone gets around to it.

Don-Ward:
Hi Clint,

I can put in a remark along the lines of “At present, there are a few places where …” to leave open the possibility of change if you’d like.

But note any change to allow large integers could break stuff. For example, the modification I made to large() (in ipl/procs/numbers.icn) last November actually depends on seq() returning an error if it is given a large integer.  I suspect that cases like that are quite rare though, and the proper way round that particular problem would be to introduce an observer function into the runtime that reliably distinguishes large integers from their smaller brethren.

Don

PR #133, opened 2020-09-26
Expanded CI coverage and debian build updates

Jafaral:
This PR includes:
* Expanded continuous integration testing. The CI now covers:
    - Ubuntu 18.04 and Ubuntu 20.04
   - Ubuntu 20.04 32-bit
    - macOS
   - Windows 64-bit 
  The CI also attempts to build Unicon on Ubuntu 20.04 with different config options:
    - No Graphics
    - No Concurrency
    - Using Clang complier
  Work to add Window 32-bit and more config options is ongoing

* Binary Package builds, currently only targeting:
   - Ubuntu 18.04
   - Ubuntu 20.04 

* PR automatic labels. When a PR is opened on GitHub, it will be labeled based on what
   files are being  changed. Example labels are IPL, iconx, and icont.

* Various fixes and updates to debian packaging.

* Silence parentheses-equality warning in clang and drop deprecated fno-unit-at-a-time flag 

* Add OpenSSL License exception text to COPYING

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,30 @@
+# Add 'repo' label to any root file changes
+repo:
+  - ./*
+
+iconx:
+  - src/runtime/*
+
+icont:
+  - src/icont/*
+
+iconc:
+  - src/iconc/*
+
+src:
+  - any: ['src/**/*']
+
+IPL:
+  - any: ['ipl/**/*']
+
+uni:
+  - any: ['uni/**/*']
+
+tests:
+  - any: ['tests/**/*']
+
+doc:
+  - any: ['doc/**/*']
+
+plugins:
+  - any: ['plugins/**/*']
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,27 +1,87 @@
 name: Build
 on:
   push:
+    branches:
+      - master
+      - ci
     paths-ignore:
     - 'docs/**'
 
   pull_request:
     branches:
       - master
+      - ci
+    paths-ignore:
+    - 'docs/**'
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
+  Windows:
+    runs-on: ${{ matrix.cfg.os }}
+    name: ${{ matrix.cfg.name }}
     strategy:
       matrix:
-        os: [macOS-latest, windows-latest, ubuntu-latest]
-
+        cfg:
+        -  { os: windows-latest,  name: 'Windows 64',   opt: '' }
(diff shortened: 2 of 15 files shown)


Don-Ward (approved):
MacOS Catalina (10.15, Kernel version 19.6) is the first OS not to run 32 bit software. Consequently, we may find a number of people stopping at its predecessor MacOS Mojave (10.14, Kernel version 18.7). Is it worth adding a build check using Mojave?

There is a missing newline at the end of debian/copyright: I don't know if it matters.

debian/rules:  We exclude stuff in the .git directory now, but the comment still refers to svn.  "custmize" is usually spelt "customize" --- for some value of usually (where I live, it's spelt customise :-)

Makefile and Makefile.in are inconsistent: Makefile.in has the @echo "did we get ..." lines removed; Makefile still has them in place. I don't know if it matters.

Jafaral:
@Don-Ward  Thanks! I fixed the issues you  pointed out.

I don't think that we have to worry about macOS 32-bit since we are at the "producing" end. We run on 64-bit and produce 64-bit code. Moreover, Apple is switching to ARM anyway soon.

Don-Ward:
Hi Jafar,

I’ve merged #133 and #134 and propagated them over to source forge.

I wasn’t suggesting we build on a 32-bit platform, I was suggesting we build on Mojave  (which is a 64 bit platform, as you point out) because it is likely that a number of Mac users will end up on Mojave — the reason being that they don’t want to “upgrade” to Catalina because they will lose the ability to run some 32-bit software — not Unicon, something else, for example an old copy of Microsoft Word — and they don’t want to do that.

Essentially, I’m foreseeing that Mojave will remain popular for some time and suggesting we might want to continue check our builds on it. For the same reason (popularity) we might want to continue checking our builds on an Intel platform when Apple moves to the new Arm processor. 

In my case I chose to put Mojave inside a VM running on Catalina, precisely so I could continue to run my elderly copies of Microsoft Office.   I did wonder if that was strictly necessary — why not try the offering from Apple?  So, just as an exercise, I thought I’d try Apple’s Excel competitor (Numbers).
      … 
After trying for three hours to get a simple scatter plot with one column acting as the X value and two other columns as the Y value with a secondary Y axis (i.e. one Y column plotted on the LH axis, the other on the RH axis) I gave up.  I’m sure it is possible to do without sacrificing your first-born but no amount of googling produced answers to what I wanted to do.

I did manage to get a single XY plot going but the secondary Y axis looks like it’s a step too far.  So it looks like Excel running inside a Mojave VM is my way forward for the foreseeable future. Life is just too short to be grappling with Apple’s spreadsheet although, if I didn’t want graphs, it does look fairly competent at the basics.

Don

PR #135, opened 2020-10-03
Fix crash on macOS if libz is enabled

Jafaral:


Don-Ward (approved):
It built for me on Catalina and Mojave with and without debug.

After investigating why the GitHub build didn't fail on commit dc18f8f1, the reason is that we build without graphics on MacOS.  We might want to fix that, if possible.

PR #137, opened 2020-10-06
Make &version a generator.

Don-Ward:
The first result is the string from version.h (exactly as before).
Subsequent results give details from the build (the output from
git describe --long HEAD) and the name of the branch that the software
was built on. The Unicon compiler uses the information to build a
more informative output when the -version option is given.


Jafaral (approved):
Looks good to me.  Once we have this in I can make further changes for packaging purposes.

Jafaral:
@cjeffery  are you OK with this change ?

PR #140, opened 2020-10-10
Add a few new config options

Jafaral:
--enable-debugheap
--enable-verifyheap
--htmldir

Also better detection of whether the C compiler really supports
Wno-parentheses-equality or not


Jafaral:
@StephenWampler 

here is  the new option: 
```
./configure --htmldir=/var/www/html/unicon
```

The result is a variable written to two files
```
Makefile
Makefile.uni
```
In your case you want  to include Makefile.uni  probably to have  the new  variable available to  you. I assume that would  be the root directory and you would build a doc tree underneath it if you need to. 

```
jafar@jtr:~/unicon$ tail -n 5 Makedefs.uni
libdir=${exec_prefix}/lib
docdir=${datarootdir}/doc/${PACKAGE_TARNAME}
mandir=${datarootdir}/man
htmldir=/var/www/html/unicon
datarootdir=${prefix}/share
```

Don-Ward (changes requested):
It probably does make sense to turn the verify-heap and debug-heap options on and off together, although they are nominally separate (and do different, but complementary, things).  

Note that the verify heap test, although enabled, won't actually do anything until an environment variable VRFY is set: -2 is probably the best value for general purposes. See rmemmgt.r circa line 2040 and init.r line 1248 for the details.  FYI; the reason it's done this way is to allow a rebuild to be possible if the Heap Verifier causes a verification crash during the system build — we need a functioning compiler to be able to rebuild. So we should set the environment variable to activate the heap verifier if --enable-verifyheap is selected.

Jafaral:
The two options are available allowing the user to turn on the ones they want if they know what they are doing. We should document all options in the "developer guide"  when that is a thing. 

I don't know what you mean bu setting the environment variable from the build perspective. The variable  should be set by the user in whatever context/shell they are running with. Ore are you just referring to to the test on Github CI ?

Don-Ward:
> The two options are available allowing the user to turn on the ones they want if they know what they are doing. We should document all options in the "developer guide" when that is a thing.
> 
> I don't know what you mean bu setting the environment variable from the build perspective. The variable should be set by the user in whatever context/shell they are running with. Ore are you just referring to to the test on Github CI ?

The latter: the test on GitHub CI should make sure that the heap verifier is active; otherwise we might think we are enabling it, but in reality nothing is happening.

Jafaral:
@Don-Ward  I pushed a change to set VRFY to -2  when  doing make Test. I wasn't sure if I want to add it to the build itself as well  to trigger the code when building IPL and Unicon libraries. I left that out.

Don-Ward:
> @Don-Ward I pushed a change to set VRFY to -2 when doing make Test. I wasn't sure if I want to add it to the build itself as well to trigger the code when building IPL and Unicon libraries. I left that out.

@Jafar It's up to you; but my preference would be to enable it when building the system itself (on just the one build where the heap verifier is enabled). It would be a much more thorough test and, even if the verifier caused a controlled crash, it wouldn't affect any other builds.

Jafaral:
I defined VRFY globally on Linux for build and test. I'm assuming if VerifyHeap is not defined then VRFY doesn't do anything, correct ?  Maybe it worth enabling  debug and verify heap on multiple of those tests anyway, but I'll leave that for another experiment.

StephenWampler:
| The result is a variable written to two files
| 
| Makefile
| Makefile.uni
| In your case you want to include Makefile.uni probably to have the new variable available to you. I assume that would be the | | | root directory and you would build a doc tree underneath it if you need to.

It would be nice if the last few lines of Makedefs.uni could be sourced from a script and not just from a Makefile. either by moving the make rules someplace else or putting those last lines in a file that could be include by both Makedefs.uni and scripts.

Otherwise, this is fine.

Jafaral:
> It would be nice if the last few lines of Makedefs.uni could be sourced from a script and not just from a Makefile. either by moving the make rules someplace else or putting those last lines in a file that could be include by both Makedefs.uni and scripts.

@StephenWampler  What kind of script? where/how are you going to invoke it ?

Don-Ward (approved):
Looks fine to me.

Jafaral:
Ok, so the scripts you are referring to are not part of Unicon itself. I think this is  out of scope of what the configure script ought to be doing but that doesn't mean we can't do it.  Such variables should be baked into the binaries themselves probably where  you can ask unicon what does it think the html pages are located. I will think about that a little more. I'm going to go ahead and merge this PR the if you have no objections @StephenWampler .

StephenWampler:
No objection!

PR #141, opened 2020-10-15
Development Version 13.2

Jafaral:
Since 13.1 was released a while back, development should be on the next release 13.2 now. A new release should trigger the following development version number from now on.


src/runtime/keyword.r
```diff
@@ -922,8 +922,8 @@ keyword{3} version
     }
    inline {
       suspend C_string Version;        /* As before, from version.h */
-      suspend C_string gitDescription; /* output from "git describe --long HEAD" */
-      return  C_string gitBranch;      /* output from "git branch --show-current" */
+      suspend C_string gitDescription; /* git describe --long HEAD */
+      return  C_string gitBranch;      /* current git branch */
```
Don-Ward, line 926: Would /* branch used to build the system */ be slightly better?

Don-Ward (approved):
This looks good.

PR #142, opened 2020-10-21
Unicon Windows installer enhancements 

Jafaral:
Allow building the installer from the command line
Add Makefile rule to build the installer
Update to use Inno Setup version 6
Enable installer build on GitHub
Add postscript to allow the installer to patchstr the binaries


cjeffery (approved):
To the extent that I can tell without trying them out, these changes look great and the new capabilities are awesome.

Don-Ward (approved):
It all looks ok.  Some of the successful builds are displaying errrors e.g.
   fatal: No names found, cannot describe anything.
   error: unknown option `abbrev-ref'
and the displayed version (and the revision field) of &features is strange
   Unicon Version 13.2.  October 15, 2020 (commit 4b9b672)
   UNIX ....
   Revision 1-4b9b672
   ....
But I don't know if that's because of these commits or another snafu; I suspect the latter.
4b9b672 is the hash of the merge commit (and shouldn't have a tag), so I'm not sure what's going wrong.

Jafaral:
@Don-Ward 
```
error: unknown option `abbrev-ref'
```
Windows git doesn't seem to have that command. I got this on my local Windows build as well. We have to fix it, but that is not related to this PR. 

The revision info show as `1` because  currently the build on github is configured with to do a shallow clone. (i,e, only clone that last revision) so we  end up with a revision history count of 1.

PR #145, opened 2020-10-27
Uni lib

brucerennie:


brucerennie:
These changes are to remove implementation dependencies with the package lang procedures. The suggested changes use the Unicon functions classname() and serial() to obtain the required information.

PR #147, opened 2020-10-27
Always use the unvarnished branch name when building.

Don-Ward:
If the branch name is the same name as a tag (which will happen on
long term maintenance branches) using git rev-parse to extract the
branch name results in extra information to make it "non-ambiguous".
Using git branch --show-current always returns the branch name with
nothing added.


Jafaral:
The problem is that --show current failed on some platforms.  You need a newer git to us it.

Jafaral (changes requested):
`--show-current` is not available in older git  which is  still common on many supported platforms.

PR #149, opened 2020-10-29
Minor updates to union.icn and scan_util.icn

StephenWampler:
In scan_util.icn, the code for moveMatch() is cleaned up and the header
comment adjusted to reflect the actual behavior.

In union.icn the decoding of class and record UniON strings is simplified and a
potential conflict arising from conflicts between user-defined class fields and
the internal class fields is addressed.

### DIFF
--- a/uni/lib/scan_util.icn
+++ b/uni/lib/scan_util.icn
@@ -156,12 +156,13 @@ class FindFirst : Object(fchars,cMaps,subs,lastMatch)
    end
 
    #<p>
-   #  Move past the match, returning it.  Fails if no match yet.
+   #  Move past the match, returning it.  Fails if no match yet or if the
+   #  scan position has been moved since the last call to <tt>locate()</tt>.
    #   This is a matching procedure.
    #  <[returns matched substring]>
    #</p>
-   method moveMatch(s,i,j)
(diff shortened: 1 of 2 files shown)


uni/lib/scan_util.icn
```diff
@@ -156,12 +156,13 @@ class FindFirst : Object(fchars,cMaps,subs,lastMatch)
...
    #<p>
-   #  Move past the match, returning it.  Fails if no match yet.
+   #  Move past the match, returning it.  Fails if no match yet or if the
+   #  scan position has been moved since the last call to <tt>locate()</tt>.
    #   This is a matching procedure.
    #  <[returns matched substring]>
    #</p>
-   method moveMatch(s,i,j)
-      return 1(\lastMatch, move(*lastMatch,s,i,j))
+   method moveMatch()
```
Jafaral, line 164: Is this a function introduced recently? It if is old and being used by other users we run the risk of breaking their code. Just making sure.

StephenWampler: If they're using this method, and the change breaks things, then they've been using it incorrectly.  The old implementation was faulty and could have led to unexpected behavior that was almost certainly not what was intended and would not have matched the behavior as described in the comments.

PR #151, opened 2020-11-01
Icon fixes

Jafaral:
I pulled a number of recent fixes from Icon. See individual commits for comments.


Jafaral:
@Don-Ward  @cjeffery ,  can I have you review these changes?

Jafaral:
Any word here @cjeffery  and @Don-Ward ?

src/common/dlrgint.c
```diff
@@ -25,10 +25,9 @@ dptr da;
 /*
  *  bignum -> real
  */
-double bigtoreal(da)
-dptr da;
+int bigtoreal(dptr da, double *result)
```
Don-Ward, line 28: It seems slightly odd to have the modern syntax for the parameter types, when the rest of the file has the "old school" way of specifying them. Especially as its counterpart (in src/runtime/rlrgint.r line 277) has also been modified, but retains the old school syntax.

src/runtime/cnv.r
```diff
@@ -96,7 +96,7 @@ double *d;
       case T_Lrgint:
          result.dword = D_Lrgint;
 	 BlkLoc(result) = (union block *)numrc.big;
-         *d = bigtoreal(&result);
+         return bigtoreal(&result, d);
```
Don-Ward, line 99: Should we delete the next line? It's unreachable.

src/runtime/fmisc.r
```diff
@@ -595,7 +595,7 @@ function{1} ishift(i,j)
          runerr(101,j);
       cj = IntVal(j);
       if (Type(i) == T_Lrgint || cj >= WordBits
-      || ((ci=(uword)IntVal(i))!=0 && cj>0 && (ci >= (1<<(WordBits-cj-1))))) {
+      || ((ci=(uword)IntVal(i))!=0 && cj>0 && (ci >= (1L<<(WordBits-cj-1))))) {
```
Don-Ward, line 598: Given we are comparing (1L <<( ...)) with an unsigned variable, I think 1uL would be more appropriate rather than relying on the compiler to correctly implement the promotion and integral conversion rules.  Either works (probably), but 1uL makes it marginally clearer.

src/runtime/oarith.r
```diff
@@ -846,11 +858,15 @@ operator{1} ^ powr(x, y)
 	 return real
 	 }
       inline {
+	 double z;
 	 if (x == 0.0 && y < 0.0)
 	     runerr(204);
 	 if (x < 0.0)
 	    runerr(206);
-	 return C_double pow(x,y);
+	 z = pow(x, y);
```
Don-Ward, line 866: I'm not sure why this code isn't using RealResult()

Don-Ward (approved):
@Jafaral Looks Ok. I made some comments on individual changes but there are no show stoppers.

src/common/dlrgint.c
Jafaral: I usually migrate to the new style when possible, i.e, if I  touch something old I update to the new style. I will fix the other one.

PR #152, opened 2020-11-05
Ipl procs

brucerennie:
I have corrected a couple of minor spelling errors in paths.icn and scan.icn and I have updated the equiv.icn to handle the case for procedures, files and windows where the two parameters refer to the same value by using the Unicon function image() to compare if they are the identically same value.

### DIFF
--- a/ipl/procs/equiv.icn
+++ b/ipl/procs/equiv.icn
@@ -91,7 +91,7 @@ procedure equiv(x1,x2,done)		#: compare values for equivalence
             if not equiv(x1[i],x2[i],done) then fail
       "table": if not equiv(sort(x1,3),sort(x2,3),done) then fail
       "set":   if not equiv(sort(x1),sort(x2),done) then fail
-      default: fail			# Vaues of other types are different.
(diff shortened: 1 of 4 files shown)


ipl/procs/equiv.icn
```diff
@@ -76,12 +86,12 @@ procedure equiv(x1,x2,done)		#: compare values for equivalence
...
 
    case code of {
-      "list" | "record": 
+      "list" | "record":
          every i := 1 to *x1 do
             if not equiv(x1[i],x2[i],done) then fail
       "table": if not equiv(sort(x1,3),sort(x2,3),done) then fail
       "set":   if not equiv(sort(x1),sort(x2),done) then fail
-      default: fail			# Vaues of other types are different. 
+      default: fail			# Vaues of other types are different.
```
Jafaral, line 94: Another typo: Vaues

Jafaral (approved):

ipl/procs/equiv.icn
brucerennie: I have added another commit to update the spelling in the comment

PR #153, opened 2020-11-11
Add minimum C compiler requirements to the top level README

Don-Ward:
Add a reference to the Icon porting guide. Clarify that the references
are to the Arizona documentation, not to the doc directory of the
Unicon repository.


cjeffery (approved):
Fine. Should remove references to remote (Icon) docs, or if the docs are useful, actually add them to our doc/ directory, which was the intention expressed here originally I think.

Jafaral:
I was actually thinking of dropping a lot of that text in the readme file anyway. See my draft here:

https://github.com/Jafaral/unicon/wiki

PR #155, opened 2020-11-20
udb fixes on Windows

Jafaral:


cjeffery (approved):
Hard to argue with the great new 64-bit logic here since it is the same as the legacy 32-bit logic.
Future opportunity: the duplicate code can be factored into a helper function.

PR #156, opened 2020-11-22
clang  warning fixes and minor udb tweaks 

Jafaral:


uni/udb/srcfile.icn
```diff
@@ -785,7 +785,10 @@ end
 #
 #
 method pathOSCompat(path)
-   return replace(path, "/", "\\")
+   if &features == "MS Windows NT" then
```
Don-Ward, line 788: Should this also happen if "MS-DOS" is in &features?

Don-Ward (approved):
Looks OK. One (minor) question about MS-DOS.

uni/udb/srcfile.icn
Jafaral: Not sure if MS-DOS is still even a thing but this test should work on all windows builds.

Don-Ward: Agreed.  The reason I asked is because in many other places that deal with paths where the string "MS Windows NT" can be found the code also contains a reference to "MS-DOS".  See tmpname() in ipl/procs/io.icn for one example; there are many others.  If we're planning to remove "MS-DOS" then fair enough; otherwise, for consistency, it might be worth adding here.

Jafaral: I don't disagree. In fact, I did look at io.icn to see if I can use one of the functions there as pathOSCompat. This seemed like a good candidate : 

https://github.com/uniconproject/unicon/blob/master/ipl/procs/io.icn#L557

but it is doing more than we need, i,e, trimming the file names to be 8 chars max to be be DOS compliant. 

We also have this at the top:
https://github.com/uniconproject/unicon/blob/master/ipl/procs/io.icn#L265

However it is not used everywhere throughout io.icn. There are several instances throughout the sources where we check for Windows but not MS-DOS. For those reasons, I decided to not worry about MS-DOS in this instance. As there are probably more  things to worry about to get MS-DOS to work if it doesn't work already. And as you can see in dosname(), there is more work to be done for DOS names anyway.  udb would be low priority.

On a serious note, we do need to document what each of those checks mean, and probably add new functions to ipl for doing things beyond dos, like winname() and so on.

PR #157, opened 2020-11-24
readme: update build instructions and drop outdated info [skip ci]

Jafaral:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,64 +1,107 @@
-Unicon 13.x
-===========
+Unicon
+======
 
-This is the Unicon distribution.  Please tell us
-where it compiles, and when and how it doesn't.
-
-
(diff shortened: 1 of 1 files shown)


Don-Ward (approved):
This supersedes PR #153, which added another Arizona document (doc/port.htm) to the now deleted list.  Clint's comment at the time was "Should remove references to remote (Icon) docs, or if the docs are useful, actually add them to our doc/ directory, which was the intention expressed here originally I think".  We haven't added anything to the docs directory which implies that we think they aren't useful. But I think the porting document _is_ useful, which is why I added it with PR #153. Please consider adding it as part of this PR.

Jafaral:
The porting document and any other document can be added, but I don't a see need to have them part of this PR. We do need developer/maintainer documentation as well.

PR #161, opened 2020-11-30
Script version and date, date is extracted from the last commit

Jafaral:
config/scripts/version.sh is responsible for versioning.
Version date is extracted from the commit date.
Tarballs must be created after running the configure script once.
Fall back to old method if everything else fails

### DIFF
--- a/config/scripts/version.sh
+++ b/config/scripts/version.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 UNICON_VERSION=13.2
+UNICON_VERSION_DATE=November 28, 2020
 
 # Interested in version only
 if [ "x$1" = "xversion"  ]; then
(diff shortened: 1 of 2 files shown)


Don-Ward (approved):
The only (minor) problem that I can foresee is that UNICON_VERSION_DATE will be produced in the "national representation", which might include Unicode.  Would that give a Unicon program attempting to parse the version string a problem?  It might be better to specify the C Locale for the git command that produces UNICON_VERSION_DATE so it remains in the same format as previously.

Jafaral:
@Don-Ward , got ideas on how to do that ?

Jafaral:
I thought  you meant  there is some command line option, but I think you meant something like:
```
LC_ALL=C git log -1 --format=%ad --date=format:'%B %d, %Y'
```

Jafaral:
@Don-Ward , I did a force push adding `LC_ALL=C`, let me know if this addresses your concern.

Don-Ward:
> I thought you meant there is some command line option, but I think you meant something like:
> 
> ```
> LC_ALL=C git log -1 --format=%ad --date=format:'%B %d, %Y'
> ```

That's exactly what I meant; sorry for any confusion.

PR #165, opened 2020-12-04
iconc: force clang to not ignore flog on macOS

Jafaral:
Inspired by: Don Ward


Don-Ward (approved):
This fix solves the "undefined _flog" problem on Catalina and Mojave (and there is no reason to suspect it won't continue to work going forward). The comments make it appear to apply only to macOS but, as has been pointed out, clang may behave similarly on non macOS platforms.

PR #166, opened 2020-12-06
Another round of iconc fixes

Jafaral:
This PR fixes some iconc issues when run from an install location. More work is needed,
but I will leave that for the next release as I think we need a better strategy to maintain
the runtime libraries needed by iconc and potentially dynamic loading C functions.

### DIFF
--- a/Makedefs.in
+++ b/Makedefs.in
@@ -101,6 +101,7 @@ WUNICONEXE=@WUNICONEXE@
 ICONCEXE=@ICONCEXE@
 
 SHELL=sh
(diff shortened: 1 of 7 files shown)


Don-Ward (commented):
I'm not sure about the cmain.c fix: relfile() is already supposed to append a trailing '/' if the second parameter has one. If it's not doing that under some circumstances, we should fix it there. Otherwise, we'll have to fix it everywhere relfile is called with a trailing '/' -- e.g. in rttmain.c.

Jafaral:
Superseded by #167

PR #167, opened 2020-12-07
Create unicon/rt subdir for runtime libraries

Jafaral:
Create unicon/rt/lib and unicon/rt/include  for  runtime libraries.
This will improve binaries mobility and fixe issues after installing Unicon to standard locations

### DIFF
--- a/src/iconc/cmain.c
+++ b/src/iconc/cmain.c
@@ -122,24 +122,11 @@ char **argv;
       }
    fprintf(stderr,"iconc library files found in %s\n",refpath);
 #else					/* ExpTools */
-   if ((int)strlen(patchpath) > 18)
-      refpath = patchpath + 18;
-#endif					/* ExpTools */
-   else
-      refpath = relfile(argv[0], "/../");
-
-   /*
-      if (refpath[strlen(refpath)] != '/') {
-       unsigned int n = (unsigned int)strlen(refpath) + 2;
-       char * refpathtmp = (char *)alloc(n);
-       strncpy(refpathtmp, refpath, n-1);
-       refpath = refpathtmp;
-       strcat(refpath, "/");
-     }
-    */
 
    // find the path to the top level directory
-   refpath = unirootfile(argv[0], "/../../");
+   refpath = unirootfile(argv[0], "/../..");
+
+#endif					/* ExpTools */
 
    /*
     * Process options.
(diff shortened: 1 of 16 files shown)


Jafaral:
Iconc seems  to work now from installed  location

Don-Ward (approved):
A brief test (on macOS Catalina) confims that what was not working after make install is now working. The alternative — of setting the path to include the in-tree bin directory, which always used to work — still works as before.

cjeffery (changes requested):
is refpath[strlen(refpath)] off by 1?  should be refpath[strlen(refpath)-1] right?

src/iconc/cmain.c
```diff
@@ -121,10 +123,23 @@ char **argv;
...
+       unsigned int n = (unsigned int)strlen(refpath) + 2;
+       char * refpathtmp = (char *)alloc(n);
+       strncpy(refpathtmp, refpath, n-1);
+       refpath = refpathtmp;
+       strcat(refpath, "/");
+     }
+    */
+
+   // find the path to the top level directory
+   refpath = unirootfile(argv[0], "/../../");
```
Jafaral, line 142: @cjeffery, when you browse the changes om github, you can click the small "plus" sign button that appears next to a line when  you hover over it to a add comment tat references that line. You can always select then drag to comment on a whole block like what I'm doing here. 

The  last line I added `refpath = unirootfile(argv[0], "/../../");` replaces all lines from 104 to 140. I don't know what ExpTools is, or if I can safely drop all of that code.

Jafaral:
> is refpath[strlen(refpath)] off by 1? should be refpath[strlen(refpath)-1] right?

you are probably right. However that is commented out code and will go away.

src/iconc/cmain.c
```diff
@@ -124,7 +124,16 @@ char **argv;
       refpath = patchpath+18;
 #endif					/* ExpTools */
    else
-     refpath = relfile(argv[0], "/../");
+      refpath = relfile(argv[0], "/../");
+
+     if (refpath[strlen(refpath)] != '/') {
```
cjeffery: If this is commented-out code that is going away, why is it listed in green with + signs to the left of it and no indication of it being commented?  Perhaps it is a user interface issue.

cjeffery: Oh, is this commented out and going away?  That is not at all clear from the view I am presented with.

Jafaral: I added it in an earlier commit, but commented it out in later commit. If you are looking at the specific commit where I added it then you are correct. I do  see the label "Outdated"  in your comment above the code when you look at the comment in the PR.

Jafaral: This is all a result of a work-in-progress changes, my apologies for the  confusion.

Jafaral:
@cjeffery I cleaned up the commits a bit to eliminate code churn between commits.  My question  regarding ExpTools and the refpath calculation there remain relevant. 

Thanks!

Don-Ward:
I did some digging (with git log -L 102,122:src/iconc/cmain.c) and these lines have been changed twice since 2001. Clint permanently suppressed the reporting of build date/time in Sep 2012 — although the modification looks conditional, it isn't — and Mike Wilder introduced the ExpTools mod in Mar 2007.  ExpTools is only used in cmain.c and appears to be a way of telling iconc where the library files are by using an environment variable. It's not documented and has been superseded.  I think it can safely be removed. The only reason for retaining it I can think of would be if we wanted iconc to use a different set of libraries to the rest of the system, perhaps during development.

Jafaral:
@Don-Ward thanks for digging the info. I will retain ExpTools but get rid of or clean up the rest of the code. I might not get around to do that for another day or two.

Jafaral:
@Don-Ward, this PR is good to go!

PR #175, opened 2020-12-19
UniDoc: set a new default rule. Build Ivib in GUI mode on Windows

Jafaral:


Jafaral:
@StephenWampler  I aligned UniDoc Makefile a bit with how we build other tools. I also add it to unicon/bin so that it gets installed as we prepare for 13.2 release. Let me know if this works for you.

PR #176, opened 2021-01-01
Allow Unicon to co-exist with Icon

Jafaral:


configure
```diff
@@ -1441,6 +1447,8 @@ Optional Features:
   --enable-verifyheap     Enable advanced heap verification
   --enable-devmode        Developper mode: turn on warnings and debugging
                           options
+  --enable-uniconx        Change then name of executables to co-exist with
```
celtic-coder, line 1450: Change *name of *the executables

configure
```diff
@@ -1441,6 +1447,8 @@ Optional Features:
   --enable-verifyheap     Enable advanced heap verification
   --enable-devmode        Developper mode: turn on warnings and debugging
```
celtic-coder, line 1448: *Developer mode

Jafaral: Thanks for catching these typos celtic-coder!

celtic-coder: You are welcome! 🙂

cjeffery (approved):
Looks swell.  Maybe in one comment "then name" should read "the name".

Don-Ward (approved):
None of these comments is a show-stopper (i.e. none of them is making a request to change
something before #176 is merged).

We might need to add the new names to the COPYING file and the debian copyright file.

The names "rtt" and "iconx" are still in widespread use, notably in the config heirarchy.
Is config used any more?  If not, perhaps now is the time to remove it, rather than
attempt to propagate the rtt -> urtt and uiconx -> uniconx changes. If removal is a step
too far, should we contemplate an "Archive" heirarchy, where old software goes to die?
Idol would also be a good candidate to treat in this way.

What about documentation? The Unicon book, surprisingly, gets away almost scot free but
the implementation book will require amendment. Is a single editorial note, informing
readers of the two names and saying when we say "iconx" we mean either "iconx" or
"uniconx", enough or will some more extensive editing be required? Does the same policy
work for the Unicon book?

rtt itself uses "rtt" in some of its error messages. I suspect we can rely on (rtt/urtt)
users not to be confused by this, so it's probably not necessary to change.

Some of the ipl needs to be modified to deal with the new names (for example, the
procedure dopen() in ipl/procs/io.icn).

The tests will need to be modified to cope with the new names.

On a system where both Icon and Unicon are installed, I still think it might be worth
investigating whether uniconx could hand off execution to iconx if it discovered an Icon
runtime version (i.e #176 goes a long vway towards enabling Unicon and Icon to live
together on the same system, but we could make it better).

Jafaral:
@Don-Ward  you raise many good points, remember though that this PR doesn't change the default names or add new names. All it does is allow the user to change they names.

Don-Ward:
> @Don-Ward you raise many good points, remember though that this PR doesn't change the default names or add new names. All it does is allow the user to change they names.

The fact that it's an option means the documentation task is easier -- we can just add an explanatory paragraph somewhere describing the optional names. For the rest of my comments (apart from the removal of "dead wood") I don't think it makes any difference that it's optional: we don't want the option, if taken, to break existing functionality so some follow-on work will be required before we release anything.

PR #185, opened 2021-01-23
Add -dd option to iyacc (to produce Java compatible definitions).

Don-Ward:
This is a patch from Clint to add the -dd option.


cjeffery (approved):
Looks good, thanks for the usage message improvements.

PR #193, opened 2021-02-11
Reduce the number of warnings produced by clang.

Don-Ward:
Most of the warnings are dealt with by addressing the complaint(s),
rather than suppressing the warning messages.  Some of the more
difficult issues have been left for a later effort but, with these
changes, the number of warnings is reduced to about one sixth of its
former value (~600 --> ~100 ).

### DIFF
--- a/src/common/mlocal.c
+++ b/src/common/mlocal.c
@@ -103,11 +103,25 @@ char *relfile(char *prog, char *mod) {
 
    canonize(buf);			/* canonize result */
 
+#if __clang__
+/*
+ * clang moans about extra parentheses [-Wparentheses-equality] when MSDOS
+ *  is not defined.  NB. MSDOS actually means "MSDOS and descendants"
+ *  -- i.e. it includes MS WINDOWS.
+ */       
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wparentheses-equality"
+#endif                    /* clang */
+
    if ((mod[strlen(mod)-1] == '/')	/* if trailing slash wanted */
 #if MSDOS
        ||(mod[strlen(mod)-1] == '\\')
 #endif
        )
+#if __clang__
+#pragma clang diagnostic pop
+#endif                    /* clang */
+
       sprintf(buf+strlen(buf), "%c", FILESEP);/* append to result */
    return salloc(buf);			/* return allocated string */
    }
--- a/src/runtime/rmisc.r
+++ b/src/runtime/rmisc.r
@@ -1843,7 +1843,7 @@ int pattern_image(union block *pe, int prev_index, dptr result,
                 image_case = PT_VP;
              else if(Blk(ep, Pelem)->pcode == PC_BreakX_MF)
                 image_case = PT_MF;
-             else if(Blk(ep, Pelem)->pcode == PC_BreakX_CS)
+             else /* if(Blk(ep, Pelem)->pcode == PC_BreakX_CS)  <-- must always be true */
                 image_case = PT_EVAL; 
 
 	     if (construct_funcimage(ep, image_case, Blk(ep, Pelem)->pcode, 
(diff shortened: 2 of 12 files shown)


src/runtime/rmisc.r
```diff
@@ -1843,7 +1843,7 @@ int pattern_image(union block *pe, int prev_index, dptr result,
                 image_case = PT_VP;
              else if(Blk(ep, Pelem)->pcode == PC_BreakX_MF)
                 image_case = PT_MF;
-             else if(Blk(ep, Pelem)->pcode == PC_BreakX_CS)
+             else /* if(Blk(ep, Pelem)->pcode == PC_BreakX_CS)  <-- must always be true */
```
Jafaral, line 1846: I don't get this.

src/common/mlocal.c
```diff
@@ -103,11 +103,16 @@ char *relfile(char *prog, char *mod) {
...
+#if __clang__
+   /* clang moans about extra parentheses [-Wparentheses-equality] */
+   if (mod[strlen(mod)-1] == '/')	/* if trailing slash wanted */
+#else                    /* clang */
    if ((mod[strlen(mod)-1] == '/')	/* if trailing slash wanted */
 #if MSDOS
        ||(mod[strlen(mod)-1] == '\\')
 #endif
        )
+#endif                    /* clang */
```
Jafaral, line 115: Why not just use -Wno-parentheses-equality, which we do anyway. The new if statement suggests that clang and MSDOD (windows?) are mutually exclusive which not the case. We need a better fix fir this one.

Jafaral (changes requested):

src/runtime/rmisc.r
Don-Ward: Here is an extract from our email conversation (Oct 2020). 

rmisc.r:1846:22: warning: variable 'image_case' is used uninitialized whenever 'if' condition  is false [-Wsometimes-uninitialized]
          case PC_BreakX_VF:
          case PC_BreakX_VP:
          case PC_BreakX_MF:
          case PC_BreakX_CS: {
             int image_case; 

             if (Blk(ep, Pelem)->pcode == PC_BreakX_VF)
                image_case = PT_VF;
             else if(Blk(ep, Pelem)->pcode == PC_BreakX_VP)
                image_case = PT_VP;
             else if(Blk(ep, Pelem)->pcode == PC_BreakX_MF)
                image_case = PT_MF;
             else if(Blk(ep, Pelem)->pcode == PC_BreakX_CS)
                image_case = PT_EVAL; 

             if (construct_funcimage(ep, image_case, Blk(ep, Pelem)->pcode, 
                             result, index_image) == RunError) return RunError;

I think clang is just being dim here. The “offending” code is inside a switch statement
      switch (Blk(ep,Pelem)->pcode) {
So, given the four case statements, image_case can never end up not being initialised. We can either suppress the warning or, perhaps, omit the final if statement like this
             else /* if(Blk(ep, Pelem)->pcode == PC_BreakX_CS) */
                image_case = PT_EVAL; 
Or we can swallow our pride and initialise the variable, even if it doesn’t actually need to be initialised.
The final possibility is 
             int image_case = (Blk(ep, Pelem)->pcode == PC_BreakX_VF)
                               ? PT_VF 
                               : ((Blk(ep, Pelem)->pcode == PC_BreakX_VP)
                                   ? PT_VP
                                   : ... etc.
But I don’t think that’s as clear as the present code.

src/common/mlocal.c
Don-Ward: I just checked Appendix H of the IB and, the comment there is 
        MSDOS    IU        build on Microsoft DOS and descendants (includes Windows, LOL)
I'll suppress the warning with a pragma (and include a reminder to the unwary reader that MSDOS includes WINDOWS).

Don-Ward:
I’ve fixed this now with a pragma (and force pushed the new version)

PR #195, opened 2021-02-23
Add compiler flag std=-c11

Jafaral:

### DIFF
--- a/src/h/opengl.h
+++ b/src/h/opengl.h
@@ -555,8 +555,7 @@ struct fontsymbol {
 //#define CWIDTH 0.25
 //#define CNEAR 0.25
 //#define CFAR 50000.0
-//#define CHEIGHT(w) \
-//  (CWIDTH*((double)w->window->height)/((double)w->window->width))
+//#define CHEIGHT(w) (CWIDTH*((double)w->window->height)/((double)w->window->width))
 
 #define CNEAR 0.25
 #define CFAR 50000.0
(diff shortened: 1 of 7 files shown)


Don-Ward (commented):
The change to src/h/opengl.h was made by commit
     223c6a6c  11-Feb-2021 | Reduce the number of warnings produced by clang.
I'm wondering why it's showing up as a diff here.

Apart from that, it looks OK.

src/h/opengl.h
```diff
@@ -555,8 +555,7 @@ struct fontsymbol {
 //#define CWIDTH 0.25
 //#define CNEAR 0.25
 //#define CFAR 50000.0
-//#define CHEIGHT(w) \
-//  (CWIDTH*((double)w->window->height)/((double)w->window->width))
+//#define CHEIGHT(w) (CWIDTH*((double)w->window->height)/((double)w->window->width))
```
Jafaral, line 558: Are you referring to this line @Don-Ward ?

Don-Ward: Yes

Jafaral: You added `//` at the front of the second line. My compiler was still complaining about "multi line comment".  I move it all to be one line. So this is a new change

Don-Ward: OK

Don-Ward (approved):

PR #196, opened 2021-02-24
Testing of the operator overload facilities: Initially against the UTF8 classes

brucerennie:
This draft pull request is for the purposes of testing the operator overload facilities that are normally turned off in the Unicon distribution. The initial testing is against the UTF8 classes which can use the operator overload facilities if available,


brucerennie:
Interesting results in that, the Ununtu 32-bit is a segmentation fault when running the non-overload code in UTF8 test
MacOS has a failure in compiling the UTF8 test code and
Thin build also a segmentation fault when running the UTF8 code and 
Windows, well we have some issues here with operator overloading. 

Result, I don't see that we can move forward with the operator overloading facilities until there are further investigations here

brucerennie:
Thanks for setting to draft. I am missing where that is set.

brucerennie:
Jafar,

I have been adding each of the new commits to both the UTF8 and the UTF8 with operator overload and checking the results of each run. The operator overload on the Unbuntu 32 bit is no longer failing with a segmentation error, even though it is failing without operator overload. Are you able to get any further information about  this?

Jafaral:
@brucerennie, GitHub has been experiencing issues  with their "Action" infrastructure in the last couple of days but they seem to have sorted this issues today. Let see how things progress in the next day or day, I can take  some actions if needed.

brucerennie:
This was a draft for testing the operator overload facilities against the UTF8 library. This is no longer needed. There is further work being undertaken to look at improving the operator overload facilities

PR #199, opened 2021-03-04
Streamline documentation build

Jafaral:
Add top level configuration rules
Add Makefile and rules to enable full doc build


Don-Ward (approved):
Now the documentation can be built with configure and make, should make clean remove the built documentation files if the system is configured with --enable-doc ?

Jafaral:
Good catch, I will add that.

PR #202, opened 2021-03-09
runtime: use stat() not lstat() to allow us follow symlinks

Jafaral:


Don-Ward (approved):
There appears to be a memory leak (of filebuffer) but, if an error occurs, the program will exit so perhaps it doesn't matter.

PR #204, opened 2021-03-15
Fix for object.icn to obtain class names with package.

brucerennie:
The Unicon function classname does not return the package that a class is in.
This affects UniDoc and causes an abort in the operation of UniDoc.

The change to object.icn in the lib has been adjusted to fix this error.

### DIFF
--- a/uni/lib/object.icn
+++ b/uni/lib/object.icn
@@ -91,7 +91,18 @@ class Object()
    # <[returns the classname for the current class in package::class format]>
    # </p>
    method className()
-      return mapPackageInt2Ext(::classname(self))
+      local cl, cn
+
+      cl := ::image(self)
+      cl := cl[8:(::find("(", cl))]
+      cn := ::classname(self)
+      if ::match(cn, cl) then {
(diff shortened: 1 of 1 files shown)


StephenWampler:
This fixes the immediate issue, but I'm not sure it's the best fix.  The original code in origin.icn was:

   return Type()

But Type() no longer matches its documentation.  It should be starting its result sequence with the (fully-qualified) name of the current class following by the (fully-qualified) names of the superclasses:

    UniDoc::UPackage
    UniDoc::UEntity
    lang::Object

Instead it just produces the names of the superclasses:

    UniDoc::UEntity
    lang::Object

omitting the name of the originating class.  I believe this is the originating error.  If that were fixed, then the method className() in object.icn could be reverted back to:

    return Type()

As it is now, the proposed solution fixes className(),, but leaves Type() broken.

StephenWampler:
Oh, I wasn't clear.  The error is in the _procedure Type_ found in predicate.icn and called by the _method Type()_ in object.icn.

PR #205, opened 2021-03-16
Update function classname() to return package as well.

brucerennie:
The original classname() function removed the package part of a class name.
This change is to include the package name as a part of the class name.

The only line actually affected is line 692. This line is removed from the source. Other changes are due to my editor automatically removing extraneous spaces at the end of any line.

This change ensures that the package information is kept when getting the class name of the object selected.

### DIFF
--- a/src/runtime/fstruct.r
+++ b/src/runtime/fstruct.r
@@ -22,7 +22,7 @@ function{1} delete(s, x[n])
             register uword hn;
             register union block **pd;
             int res, argc;
-	    
+
 	    MUTEX_LOCKBLK_CONTROLLED(BlkD(s, Set), "delete(): lock set");
 	    for (argc = 0; argc < n; argc++) {
 	       hn = hash(x+argc);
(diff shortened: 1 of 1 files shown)


StephenWampler:
This looks good!  I do note, of course, that it returns "PACKAGE__CLASS" instead of "PACKAGE::CLASS".  I'm assuming that's planned as a later mod.

With this change, and changing line 50 of ${UNICON}/uni/lib/predicat.icn to:

      suspend lang::mapPackageInt2Ext(::classname(var) | !get_class(var).get_supers())

then UniDoc runs correctly.

brucerennie:
As for the __ to :: change, as I am not a c programmer and I, as yet, do not know the extent to where these changes are to be applied, this specific change is definitely for a later mod.

StephenWampler:
Also note that, with this change, the code for the className() method in ${UNICON}/uni/lib/object.icn can be reverted back to just:

      # <p>
      # <[returns the classname for the current class in package::class format]>
      # </p>
     method className()
         return lang::mapPackageInt2Ext(::classname(self))       
     end

StephenWampler:
> As for the __ to :: change, as I am not a c programmer and I, as yet, do not know the extent to where these changes are to be applied, this specific change is definitely for a later mod.

Sounds good to me.  I prefer one step at a time anyway.

StephenWampler (approved):
Looks good!  Once this is in, I'll make the changes to predicat.icn and object.icn and, once those are merged, work on updating UniDoc with the improvements Jafar mentioned.

Jafaral:
Looking forward to your fixes @StephenWampler .

Thanks @brucerennie

PR #206, opened 2021-03-17
Update to use repaired classname()

StephenWampler:
Simplify className() and Type() methods now that the classname procedure returns fully-qualified class names.

Signed-off-by: Steve Wampler <sbw@tapestry.tucson.az.us


StephenWampler:
This also fixes the Type() method to include the name of the current class in its result sequence.

PR #208, opened 2021-03-17
Update to Args class and UniDoc option handling.

StephenWampler:
This adds the ability to detect bad command line options to the util::Args class and adds that detection to UniDoc.  A bug in UniDoc where the handling of a default target directory is incorrect has also been fixed.

### DIFF
--- a/uni/unidoc/UniDoc.icn
+++ b/uni/unidoc/UniDoc.icn
@@ -132,7 +132,6 @@ import lang
 import UniDoc
 
 global linkPath,  # Path to search when looking for already-generated HTML
-       targetDir, # Directory into which all generated HTML is placed
        title,     # Title for base document
        indexFile, # Name of base index file
        debug      # Debug flag
@@ -148,7 +147,7 @@ procedure main(args)
     processArgs(uniAll)
     uniAll.dumpStatistics("yes")
 
-    html := UniHTML(title, uniAll, targetDir, indexFile, linkPath, debug)
+    html := UniHTML(title, uniAll, indexFile, linkPath, debug)
     html.buildPages()
 
 end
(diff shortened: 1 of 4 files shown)


uni/unidoc/UniDoc.icn
```diff
@@ -167,7 +168,7 @@ procedure processArgs(uniAll # Unicon parser (instance of <tt>UniAll</tt>)
     linkPath  := []	# Default situation (no --linkPath option)
     linkPath  := [: genFields(\(Args().getOptDef("linkPath")),",") :]
     uniAll.setSourcePath(Args().getOptDef("sourcePath"))
-    uniAll.setTargetDir(targetDir := Args().getOptDef("targetDir"))
+    uniAll.setTargetDir(targetDir := Args().getOptDef("targetDir","./htmlDoc"))
```
Jafaral, line 171: Since targetDir is already set to `./htmlDoc` at the top of this function I suggest just using that here.  Or maybe just delete the line since this change will always overwrite targetDir.

StephenWampler: I've made a slightly different change.  There is no longer a global variable targetDir and all references to the value are through UniAll's (net) getTargetDir() call.  I'll put that into the same PR as soon as I remember how to do that.  (Besides, I can't find where targetDir is preset at the top of that function!.

It looks like the build (see below) is failing again because fixName() isn't being found.  I don't see that here yet.

StephenWampler: Found the missing line.  I had removed it because it just gets in the way without adding any value.

Jafaral: here, but it looks like  you have deleted it in the new commit
```
    local arg
    indexFile := "index.html"
    targetDir := "./htmlDoc"
```

PR #209, opened 2021-03-18
fix upexample plugin, add test example

Jafaral:


Don-Ward (approved):
@Jafaral Looks good. Is it worth considering making the test directory (i.e. plugins/src/myPlugin/test) a standard location and adding a test target to the example makefile?  We could make it clear that implementing the test target is optional.

PR #213, opened 2021-03-21
tests: allow tests to run when unicon is not on the path

Jafaral:


brucerennie:
I cannot see any problems with these changes

brucerennie (approved):

PR #215, opened 2021-03-22
tests: re-enable some Windows special cases

Jafaral:


brucerennie (approved):
Jafar, as far as I can tell the changes you are proposing should be okay. Nothing has popped out at me with what you propose.

PR #219, opened 2021-03-29
config: add werror  and warnings options

Jafaral:


src/h/config.h
```diff
@@ -548,6 +548,8 @@
 #endif					/* IntBits */
 
 #if (WordBits == 64)
+   #define LINTFRMT "l"
```
Don-Ward, line 551: My first thought on seeing this line was "What's lint got to do with it?".  It was only later when looking at rmisc.r that I realised it meant "Long Integer".  C99 introduced the inttypes.h header with predefined printf format specifiers (PRId64 etc.)  Is it worth considering using that?

src/runtime/rwinsys.r
```diff
@@ -42,6 +42,7 @@ void checkpollevent(){
 
 #else					/* Graphics */
 #if NT
+int iconx(int argc, char **argv);
```
Don-Ward, line 45: I don't think we'll get a compiler fussy enough to complain about the difference between char ** and char*[], but why take the risk? i.e. declare it as int iconx(int argc, char *argv[]);

Don-Ward (approved):
Looks good. Perhaps my suggestion to use intypes.h is better taken as a separate item -- it would lead to quite a few changes if done systematically.

src/h/config.h
Jafaral: Given your comment below, no action is needed here then.

src/runtime/rwinsys.r
Jafaral: Only on Windows I got this warning, and I copied the function prototype verbatim from imain.r as it is actually defined.

Jafaral:
@Don-Ward , if you are OK with it then let us get in please.

PR #221, opened 2021-03-29
CI: add CentOS

Jafaral:


Jafaral:
This CI tested was incorporated in #220 which is were the problem was reported. Closing this PR.

PR #222, opened 2021-04-01
ci: refactor and add new platforms

Jafaral:
Simplify the CI script a bit, add Fedora 33 and Debian 10 to the test


Jafaral:
testing testing and more testing @Don-Ward

Don-Ward (approved):
It looks OK to me. I have one comment regarding the macOS testing we do: Currently we test on Catalina. Big Sur is the latest version and has been around long enough to perhaps justify switching our tests to using it.  We might also consider adding a test for Mojave -- which is the last macOS to support 32 bit applications (and so some folk miight wish to continue using that, rather than update it and lose access to some things they still want to run).  I guess most people will have switched to BIg Sur from Catalina by now but there may be a significant minority sticking with Mojave, so my suggestion is to test both.

PR #225, opened 2021-04-04
Upl utf8 split test into multiple short procedures

brucerennie:
Both of the UTF8 and UTF8 Operator overload test files have been split into multiple short procedures to see if this facilitates the macOS testing to NOT fail.


brucerennie:
Have updated the PR 183 and PR 196 to have the new utf8 test files. This PR can be removed

PR #228, opened 2021-04-08
Make auto.h inclusion idempotent.

Don-Ward:
Custom configuration decisions are usually made via src/h/define.h
But, if the customisation is dependent on the detected configuration,
then access is required to the definitions in auto.h. Unfortunately,
auto.h is normally included __after__ define.h. Making the inclusion
idempotetent means that define.h may include auto.h if it is required.

This also allows define.h to override the automatic configuration if
it is necessary to do so in some circumstances.

### DIFF
--- a/src/h/auto.in
+++ b/src/h/auto.in
@@ -1,4 +1,6 @@
 /* src/h/auto.in.  Generated from configure.ac by autoheader.  */
+#ifndef __AUTO_H
+#define __AUTO_H
 
 /* Define if building universal (internal helper macro) */
 #undef AC_APPLE_UNIVERSAL_BUILD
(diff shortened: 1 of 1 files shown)


Jafaral (changes requested):
This will not work. auto.in is "owned" and automatically generated by autoconf, it shouldn't be touched by hand.

auto.h and define.h should not depend on predefined symbols.

Don-Ward:
> This will not work. auto.in is "owned" and automatically generated by autoconf, it shouldn't be touched by hand.
> 
> auto.h and define.h should not depend on predefined symbols.

@Jafaral I don't believe auto.in is generated by autoconf. I'll discuss this with you privately.

PR #230, opened 2021-04-18
doc: build full html using unidoc

Jafaral:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,12 +195,27 @@ docrule: $(UDOC)
 doc:
 		$(MAKE) -C doc
 
+# build full html documentation using unidoc
+UBASE=$(shell pwd)
+SBASE=$(UBASE)/uni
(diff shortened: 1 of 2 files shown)


Jafaral:
@StephenWampler , I adapted your script to the makefile and  removed some path dependencies (relpath, /opt, etc.). It works, but hope we can make the resulting html files a bit more organized, maybe a hierarchy that reflects the  hierarchy of the source dir/files.

StephenWampler (approved):
This looks fine.  I like the idea of reflecting the source hierarchies but it will need some thought since it may be difficult to always distinguish separate hierarchies cleanly (e.g. $UBASE/progs vs $SBASE/progs).  Whatever's done needs to work with arbitrary user source directories.

Jafaral:
Here it is live @StephenWampler :

http://unicon.org/api/uni-api/

PR #231, opened 2021-04-20
Adding a new keyord &proc so that access to the underlying function proc is always available

brucerennie:
Since it is possible to shadow the Unicon function proc, I have added a new keyword &proc that returns this internal function so that no matter what the programmer may do in terms of shadowing any of the internal Unicon functions, there is always a way to access the proc function which then allows access to all other internal Unicon functions.


Jafaral:
I'd like  &proc to be added as part of PR that introduces a few other new pending keywords, since adding a keyword requires pumping up the icode version. Might as well do it in one fell swoop!  

Any thoughts @cjeffery ?

PR #238, opened 2021-07-02
Formatting changes and code examples updates/corrections to the Unicon Book

brucerennie:
I have create a number of corrections for oddities in formatting in the
current Unicon book. There are also changes to some of the tables (eg
Error Number/Error Descriptions and Preprocessor Symbols/Decriptions by
Mkaing these tables instead of the direct formatting of the lines.


brucerennie:
I should note that these changes are from a personal reading perspective and my not suit others.

cjeffery (changes requested):
Things I like here:
* replacing "manual tables" with proper LaTeX tabular environments and such
* replacing "manual indentation with proper use of \> in source code environments
* fixes for gratuitous linebreaks that make the LaTeX less readable and serve no purpose
Things I don't like here:
* loss of readability of the .tex for source code when spaces are deleted
* gratuitous changes that don't really improve LaTeX or output appearance
* introduction of lines longer than 80 columns when they are not necessary
* introduction of vertical blank lines that make code less likely to fit on a page

brucerennie:
Thanks for the feedback. 

As far as the extra spaces to improve readability in the source code, that is an easy fix and I can undertake that in the near future.
As for the lines longer than 80 columns, that was simply for my benefit while updating the latex code and is also an easy fix.
As for the additional blank lines in the code, this can be easily removed and I can do that while adding the spaces for code readability.

I do have one question and that relates to the "gratuitous changes" comment, do you have any specific examples that highlight what you mean here. Just to note: I have considered removing this draft pull request and resubmitting what I am currently up to. So, if you are able to highlight the particular kind of Latex that I have used which hasn't been an improvement or hasn't improve output appearance, I would really appreciate that so this can be corrected before uploading the current work.

brucerennie:
Due to a change in direction and trying out the listings package to address some of Clinton's concerns relating to the formatting of the various code examples, I am removing this pull request.

PR #241, opened 2021-09-14
Fix listener sockets

Jafaral:
When IPv6 support was added I was passing `is_udp_or_listener` to mean udp. It is only upd 
 if `is_udp_or_listener == 1`

I tested a tcp server with listener socket and it works now.

### DIFF
--- a/src/runtime/fsys.r
+++ b/src/runtime/fsys.r
@@ -971,7 +971,7 @@ Deliberate Syntax Error
 #endif					/* Graphics || Messaging || ISQL */
 	       /* connect to a port */
       	       DEC_NARTHREADS;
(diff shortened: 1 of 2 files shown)


Don-Ward (approved):
It looks OK.  I have two comments, both cosmetic.

1) in most cases we say "is_udp_or_listener == <value>". In one case (rposix.r #1273) we say "<value> == is_udp_or_listener", which jars slightly.
2) The GitHub listing is indenting the source incorrectly because (I'm guessing) it doesn't have tabs set to 8 spaces. Is it possible to change that?  Even better would be to not have tabs in any of our source files, but that might be a controversial suggestion.

PR #243, opened 2021-09-16
Allow &input to be used with select() and ready()

Jafaral:


cjeffery (approved):
Looks great so far as I can tell using a web interface.

PR #244, opened 2021-09-17
Allow select() to accept a set() as an argument

Jafaral:
select() accepts files or a list of files, but it is useful
to also have support for a set of files. When managing a large
number of sockets it is convenient to maintain the sockets in a set
since a socket can be easily found using member() instead of looping
through a list.

### DIFF
--- a/src/runtime/fxposix.ri
+++ b/src/runtime/fxposix.ri
@@ -2357,6 +2357,7 @@ function{0,1} select(files[nargs])
       tended struct descrip f;
       tended struct b_list *lws = NULL;
       tended struct b_list *lps = NULL;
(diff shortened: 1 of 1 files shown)


Jafaral:
@cjeffery , for context, this is mostly triggered by me having to right code like the following where I have to find and delete a socket from the list passed to `select()`:
```
             every i:= 1 to *sock_list do
                if sock_list[i] === s then delete(sock_list, i) & break
                   write("closing socket: ", image(s), "\tactive sockets: ", *sock_list)
             close(s)
```

I thought a set() would be a good fit for such use case, now I only have to do:
```
delete(sock_list, s)
```

PR #245, opened 2021-09-25
Fix socket accept mode

Jafaral:
When calling open() n mode "na", we ask for a TCP socket but without
going through an explicit listen(). In that mode, the call to open()
should be blocking, unlick a listener "nl"  mode. In "na" case we
still want to call listen() first, then invoke acceptto block.


Don-Ward (approved):
Cosmetic suggestion: add
    /* No need to listen on UDP sockets */
just before line 1337

Jafaral:
comment addressed, thanks @Don-Ward

PR #249, opened 2021-10-01
Fix behavior of sort, sortf, min, and max with respect to numerics.

StephenWampler:
Modifies the behavior of sort(), sortf(), min(), and max() so they treat integer and real values collectively as numeric values instead of separate types.  The code includes support for adding new functions to support the old behavior [just with new function names] if desired.


StephenWampler:
Pulled back because of plans to add an Icon compatibility library means this code can be greatly simplified.

PR #251, opened 2021-10-05
Fix sort(), sortf(), min(), and max() so comparisons treat integer an…

StephenWampler:
…d real values collectively instead of as separate types

This changes the comparison behavior so integers (including big integers) and reals are compared against each other by value instead of separating reals into a separate type.


Don-Ward (approved):
I built it and it works as advertised.

What the PR does not do is fix the larger issues with max() and min() when comparing non numerical types. For example
```
write(type(min(write,1)))
write(type(max(write,1)))
```
produces
```
  integer
  procedure
```
Arguably, they should both fail.  This isn't a criticism of PR251, just a note that there is more work to do in this area, especially when considering strings.

PR #260, opened 2021-12-16
Re-enable sha

Don-Ward:
The first commit puts in "the plumbing" for cnv_c_str on Windows.
The second commit enables the SHA plugin.

### DIFF
--- a/ipl/cfuncs/icall.h
+++ b/ipl/cfuncs/icall.h
@@ -467,7 +467,7 @@ do {if (sizeof(a[0]) != sizeof(double))  FailCode(102); \
 #define ExternAddr(d) ((void *)&(((struct b_external *)(d).vword.bptr)->exdata[0]))
 
 /* Useful when calling malloc() to get enough space for the block header */
-#define ExtHdrSize  ((int)&(((struct b_external *)(0))->exdata[0]))
+#define ExtHdrSize  ((uintptr_t)&(((struct b_external *)(0))->exdata[0]))
 
 /*
   2. Allocations of Icon structures.
--- a/ipl/cfuncs/icall.h
+++ b/ipl/cfuncs/icall.h
@@ -9,7 +9,7 @@
 #                 Clinton Jeffery, Jafar Al-Gharaibeh,
 #                 Don Ward
 #
-#       Date:     September 2nd 2020
+#       Date:     December 16th 2021
 #
 ############################################################################
 #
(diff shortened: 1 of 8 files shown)


ipl/cfuncs/icall.h
```diff
@@ -170,6 +174,7 @@
 #define Type(d) ((int)((d).dword & TypeMask))
 
 typedef long word;
+typedef unsigned long uword;
```
Jafaral, line 177: The way we are handling this type (size) has the potential of breaking things in unexpected ways on some platforms. We probably should be consistent with what the  runtime expect:

https://github.com/uniconproject/unicon/blob/master/src/h/typedefs.h#L18

Jafaral (commented):
Looks good to me, but we need to fix the word/uword otherwise I know that wont be working in Windows at least.

plugins/src/SHA/RFC6234/makefile
```diff
@@ -1,4 +1,6 @@
 #CFLAGS = -D ORIGINAL_SHA_CODE_NOT_THREAD_SAFE -D USE_32BIT_ONLY
+CFLAGS = -fPIC -O2
+CC = gcc
```
Jafaral, line 3: Why are we hard-coding these ? Why not including the pulgin src make file ?

Jafaral (changes requested):
makefile should  renamed to Makefile to be consist with other make files in Unicon.

Don-Ward:
Hi Jafar,

Inline ..


> In plugins/src/SHA/RFC6234/makefile <https://github.com/uniconproject/unicon/pull/260#discussion_r770713062>:
> > @@ -1,4 +1,6 @@
>  #CFLAGS = -D ORIGINAL_SHA_CODE_NOT_THREAD_SAFE -D USE_32BIT_ONLY
> +CFLAGS = -fPIC -O2
> +CC = gcc
> Why are we hard-coding these ? Why not including the pulgin src make file ?
> 
At present, the RFC6234 files do not compile with -std=c11. I have yet to investigate why not (I was more interested in getting the SHA plugin to build at all).

> In ipl/cfuncs/icall.h <https://github.com/uniconproject/unicon/pull/260#discussion_r770702843>:
> 
> > @@ -170,6 +174,7 @@
>  #define Type(d) ((int)((d).dword & TypeMask))
>  
>  typedef long word;
> +typedef unsigned long uword;
> The way we are handling this type (size) has the potential of breaking things in unexpected ways on some platforms. We probably should be consistent with what the runtime expect:
> 
> https://github.com/uniconproject/unicon/blob/master/src/h/typedefs.h#L18 <https://github.com/uniconproject/unicon/blob/master/src/h/typedefs.h#L18>This isn’t a new problem, it was already there with the existing definition of word in icall.h

I’m not quite sure what you want to do — I’m guessing that you want to reflect the definitions in typedefs.h in call.h (both versions). The question is why that wasn’t done in the first place with the definition of word in icall.h. Do we have access to IntBits?

See
ca14d877  08-Feb-2017 | Another round of plugins infrastructure [Jafar Al-Gharaibeh]
which is where the definition was introduced.

Don

PR #262, opened 2021-12-27
Fix loadfunc() to search the installation directory for plugins.

Don-Ward:
The plugin library (/usr/local/lib/unicon/plugins/lib) is searched if
the specified library has not already been located.

### DIFF
--- a/src/runtime/fload.r
+++ b/src/runtime/fload.r
@@ -96,9 +96,10 @@ function{0,1} loadfunc(filename,funcname)
          }
 
       /*
-       * If not found, look in our plugins/lib directroty
-       * We are assuming that the plugins directory is one
-       * level up from iconx which should be on the path.
+       * If not found, look in our plugins/lib directory
+       * We are assuming that the plugins directory is either one
+       * level up from iconx (which should be on the path) or it
+       * is in the installed location.
        * TODO: this code is used at least in one other place
        *       (keyword.r), so it should be pulled into a util
        *       function.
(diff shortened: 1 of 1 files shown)


src/runtime/fload.r
```diff
@@ -109,8 +110,14 @@ function{0,1} loadfunc(filename,funcname)
 	 char path[MaxPath];
 	 if (findonpath(UNICONX, path, MaxPath)) {
 	    int n = strlen(path);
+            /* Try the "not installed" location */
 	    snprintf(path+n-strlen(UNICONX), MaxPath - n, "../plugins/lib/%s", filename );
 	    handle = dlopen(path, RTLD_LAZY);	/* get the handle */
+            if (!handle) {
+              /* Try the "installed" location */
```
Jafaral, line 117: When installed, we should have a "Libraries at" location that would be a better fit for this purpose -  that is why we have "Libraries at"  after. I checked at the generated config on my mac, I have these:

```
/* bindir */
#define config_bindir "/usr/local/bin"

/* docdir */
#define config_docdir "/usr/local/share/doc/unicon"

/* exec_prefix */
#define config_exec_prefix "/usr/local"

/* libdir */
#define config_libdir "/usr/local/lib"

/* mandir */
#define config_man "/usr/local/share/man"

/* prefix */
#define config_prefix "/usr/local"
```
We need to consolidate our install script and code to better honor these values. Maybe Also add a plugin path.

Don-Ward: I'm happy to withdraw the PR in favour of you implementing a "better solution" along the lines you suggest. Alternatively, we could merge this as a stopgap (because it is an improvement over the existing situation, where plugins don't work if Unicon has been installed) until the better solution actually arrives.

PR #265, opened 2022-01-03
Initial version of the SQLite interface plugin.

Don-Ward:
The plugin is not built If the sqlite3 files that it depends
on (sqlite3.h and libsqlite3.so) are not found. The check is done
when building the Unicon system.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -545,6 +545,19 @@ AC_FUNC_FORK()
 AC_CHECK_LIB(dbm,main)
 AC_CHECK_LIB(ndbm,main)
 
+# we don't need sqlite to be added to $LIBS, not yet at least
+saved_libs="$LIBS"
+AC_CHECK_HEADERS([sqlite3.h])
+AC_CHECK_LIB(sqlite3, sqlite3_config)
+LIBS="$saved_libs"
+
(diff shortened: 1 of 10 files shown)


configure.ac
```diff
@@ -545,6 +545,17 @@ AC_FUNC_FORK()
...
+saved_libs="$LIBS"
+AC_CHECK_HEADERS([sqlite3.h])
+AC_CHECK_LIB(sqlite3, sqlite3_config)
+LIBS="$saved_libs"
+
+if test "x$ac_cv_header_sqlite3_h" = "xyes" -a  "x$ac_cv_lib_sqlite3_sqlite3_config" = "xyes"; then
+  SQLITE="sqlite"
+  AC_SUBST(SQLITE)
+fi
+
```
Jafaral, line 558: Looking at this again, we need to make the variable substitution always: 

```
if test "x$ac_cv_header_sqlite3_h" = "xyes" -a  "x$ac_cv_lib_sqlite3_sqlite3_config" = "xyes"; then
  SQLITE="sqlite"
else
  SQLITE=""
fi
AC_SUBST(SQLITE)
```

Jafaral (changes requested):

configure.ac
Don-Ward: OK, I'll update configure.ac and configure and repush.

Jafaral (approved):

PR #270, opened 2022-01-05
Document that key() enumerates the key values in DBM Database files.

Don-Ward:


doc/book/langref.tex
```diff
@@ -2586,6 +2586,7 @@ \subsection*{Built-in functions}
 table \texttt{T}. \texttt{key(L)} generates the indices from 1 to
 \texttt{*L} in list \texttt{L}. \texttt{key(R)} generates the string
 field names of record \texttt{R}.
+\texttt{key(D)} generates the string key values of a DBM database \texttt{D}.
```
Jafaral, line 2589: for an http socket, `key()` also generates the header field names :-)

Jafaral: https://github.com/uniconproject/unicon/blob/master/uni/progs/uget.icn#L116

Don-Ward: Suggest we do a separate commit for that.

Jafaral: It is all about `key()`, I don't see a reason to keep it separate

cjeffery: My Swiss professor in my Calc 3 class said "this is for here and now, not later and never".

PR #272, opened 2022-01-11
Improve treatment of no output from a piped command.

Don-Ward:
Handle failure of popen(). Distinguish between a successful command that
produces no output and a non-zero return status from the command.
Document new (and not so new) error numbers in Appendix A.


Jafaral (approved):
LGTM

PR #274, opened 2022-01-30
runtime: don't count a sleeping as a running thread

Jafaral:

### DIFF
--- a/Makedefs.in
+++ b/Makedefs.in
@@ -16,8 +16,8 @@ CC = @CC@
 
 RTDIR=$(TOPDIR)/rt
 
-CFLAGS = @CFLAGS@ -I../../rt/include
-LDFLAGS= -L../../rt/lib @LDFLAGS@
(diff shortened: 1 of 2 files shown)


Jafaral:
@Don-Ward this should fix the GC delay you reported to me. Please check if it actually works.

Don-Ward:
I tested this on macOS, where I first discovered the problem, and it works now. 
Are we still maintaining the code inside `#ifdef Macintosh`? If so, perhaps we should add something similar around line 510 of `rsys.r`. If not, maybe we should take it all out (as a separate activity).
The description in the Unicon book says "`delay(i)` pauses the program for at least `i `milliseconds": This is probably a fossil from Icon and we should change "program" to "current thread".

Jafaral:
I left Macintosh out intentionally because I think it is legacy code. However, it doesn't hurt to handle it to be consistent with other cases for now. I will push an update

Don-Ward (approved):
Looks good to me. I'll alter the book as a separate commit.

PR #278, opened 2022-02-13
Add Thread-local storage.

Don-Ward:
Add a new TLS procedure to the threads package. Write about it in the Unicon
book and document that serial() returns the serial number of the current thread.

### DIFF
--- a/doc/book/langref.tex
+++ b/doc/book/langref.tex
@@ -3255,7 +3255,8 @@ occurs if the value to be generated is a large integer.
 \noindent
 \index{serial(x)}\texttt{serial(x)} returns the serial number for
 structure \texttt{x}, if it has one. Serial numbers uniquely identify
-structure values.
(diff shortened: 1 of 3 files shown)


doc/book/langref.tex
```diff
@@ -3255,7 +3255,8 @@ \subsection*{Built-in functions}
 \noindent
 \index{serial(x)}\texttt{serial(x)} returns the serial number for
 structure \texttt{x}, if it has one. Serial numbers uniquely identify
-structure values.
+structure values. \texttt{serial()} returns the serial number of the
+current thread.
```
Jafaral, line 3259: current thread or co-expression

Jafaral (changes requested):
minor commit, looks good otherwise.

Don-Ward:
Closed in favour of PR#259, which is from the correct branch (and has the changes requested here).

PR #281, opened 2022-05-24
build: drop legacy makefile rules

Jafaral:
These don't work anymore, and people still try to use them sometimes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,21 +50,17 @@ config: configure
 
 help:
 	@echo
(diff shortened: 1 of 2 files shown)


Don-Ward:
The help target of the makefile says that people should use the targets that are being removed; so the PR needs a bit more work. The help target also mentions some non existent targets NT-Configure-GCC, W-Configure-GCC and WUnicon.

Don-Ward (approved):
Simpler is better: looks OK.

PR #285, opened 2022-10-03
Add a new implementation of morse() plus a reverse function.

Don-Ward:


uni/lib/remorse.icn
```diff
@@ -0,0 +1,157 @@
...
+#
+#	Subject:  Procedures to convert a string to and from Morse code
+#
+#	Author:	  Don Ward
+#
+#	Date:	  October 2022
+#
+############################################################################
+#
+# This is a re-implementation of morse.icn which, compared to its predecessor:
```
Jafaral, line 13: should we add the full path to morse.icn (or say IPL version)  to let the reader know which file you are referring to

uni/lib/remorse.icn
```diff
@@ -0,0 +1,157 @@
...
+#	  is easier to understand and extend.
+#
+############################################################################
+
+package remorse
+
+# ----------------------------------------------------------------------
+# translate s into Morse code
+# For compatibility with morse(), words are separated by five spaces.
+procedure Morse(s)
```
Jafaral, line 27: It is not common to capitalize library function names. Since this is in a package now, we can just do morse().

Jafaral (changes requested):

uni/lib/remorse.icn
Don-Ward: Modified as requested. Morse() was originally capitalised as it's the name of a person (Samuel Morse) but I'm happy with morse().

PR #289, opened 2022-11-10
New tojson() function that writes json without encoding unicon types

Jafaral:
`utoj()` encodes unicon types when it writes json data.
Example:

if we feed the following json data to `jtou()`:

```
{"name":"Joe","age":42,"gender":"Male"}
```

The tabel in memory:
```
T2 := table(&null)
   T2["age"] := 42
   T2["gender"] := "Male"
   T2["name"] := "Joe"
```
`utoj()` produces it as:
```
{"__unitable__":1,"gender":"Male","name":"Joe","age":42}
```

whereas the new `tojson()` function produces the original json data as:
```
{"gender":"Male","name":"Joe","age":42}
```

`utoj()` is good to encode Unicon data structures to be decoded properly
later in another Unicon program.
`tojson()` cares about json data only, not about Unicon meta data.

### DIFF
--- a/uni/lib/json.icn
+++ b/uni/lib/json.icn
@@ -10,6 +10,9 @@
 #
 # Error handling object
 #
+
+package json
+
 class ErrorHandler(filename, 
                    lineno, 
                    error,
@@ -238,12 +241,30 @@ procedure utoj(x,strict,error)
 end
 
 #
+# Same as utoj(), except that this function only produces standard JSON data
+# without adding Unicon data structure meta data
+#
+procedure tojson(x,strict,error)
+   local j;
+   jerror := ErrorHandler(error,strict)
+
+   if j := _utoj(x,jerror,"r") then
+      return j
+
+   jerror.get_err()
+end
+
+
+#
 # Given a Unicon structure, produce a JSON equivalent if possible.
+#
+# If raw is not &null, then a standard JSON data is produced without
+# adding Unicon data structure meta data
 #
 # TODO: Check for fails that need error messages
 #
-procedure _utoj(u,jerror)
-   local s, tmp
+procedure _utoj(u,jerror, raw)
+   local s, tmp, first
    case type(u) of {
       "null": return type(u)
       "string": {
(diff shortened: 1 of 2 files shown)


cjeffery (changes requested):
The line 271 in json.icn
                         if *u > 0 then s ||:= _utoj(u[1],jerror, rw)
should say raw instead of rw, right?

uni/lib/json.icn
```diff
@@ -254,38 +268,50 @@ procedure _utoj(u,jerror)
       "integer" | "real": return image(u)
       "list": {
          s := "["
-         if *u > 0 then s ||:= _utoj(u[1],jerror)
+         if *u > 0 then s ||:= _utoj(u[1],jerror, rw)
```
Don-Ward, line 271: rw should be raw, probably.

uni/lib/json.icn
```diff
@@ -237,12 +240,23 @@ procedure utoj(x,strict,error)
    jerror.get_err()
 end
 
+procedure tojson(x,strict,error)
```
Don-Ward, line 243: Suggest adding a comment: eg. Produce a JSON string without added Unicon descriptors (i.e. standard JSON).

uni/lib/json.icn
```diff
@@ -237,12 +240,23 @@ procedure utoj(x,strict,error)
...
+end
+
+
 #
 # Given a Unicon structure, produce a JSON equivalent if possible.
 #
 # TODO: Check for fails that need error messages
 #
-procedure _utoj(u,jerror)
+procedure _utoj(u,jerror, raw)
```
Don-Ward, line 259: Suggest adding to comment describing the raw parameter. eg.
if raw is absent or null then Unicon descriptor strings will be added to the JSON string. To produce standard JSON without those descriptors, specify a non-null value for raw.

Don-Ward (changes requested):
One typo -- must fix.
Two suggestions -- optional.

Jafaral:
@cjeffery and @Don-Ward , fixed the issues. I also added logic to handle all unicon types not just tables. I.e, a record in Unicon should also produce a JSON object exactly like a table does/

Beware, I also moved json to be a package! so to use json, one should do `import json` rather than `link json`

Jafaral:
record example 
```
record person(name,age,male)
u:= person("Joe", "42", "Male")
write(ximage(u), "\n")
write(tojson(u, "strict"))
```
Output:
```
R_person_1 := person()
   R_person_1.name := "Joe"
   R_person_1.age := "42"
   R_person_1.male := "Male"

{"name":"Joe","age":"42","male":"Male"}
```

PR #291, opened 2022-11-13
lib: add json schema validation

Jafaral:


tests/lib/jsontests.icn
```diff
@@ -243,3 +246,48 @@ procedure objects_equal(x,y)
...
+    } _
+  } _
+}"
+
+
+   v := jss_validate(good, schema)
+   write("schema test valid:", if v=="" then  "SUCCESS" else "FAILURE " )
+
+   v := jss_validate(bad, schema)
+   write("schema test invalid:", if v~=="" then  "SUCCESS" else "FAILURE " )
```
Don-Ward, line 291: The helper function called by jss_validate has a large number of cases which are not explored by these two tests. Would it be possible to increase the coverage?  I'm not asking for 100% (or anywhere near) but this exercises only a tiny fraction of the code.

uni/lib/json.icn
```diff
@@ -221,15 +221,16 @@ procedure jtouf(s,strict,error)
    jerror.get_err()
 end
 
-#
-# It is insane to propose a single function as the primary API for the
```
Don-Ward, line 225: I quite liked the self deprecating nature of this comment but, if it has to go then so be it.

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+#<p>
+# validate a json file using a json sechema
+# <[param jfname - the name of the json File]>
+# <[param sfname - the name of the json Schema File]>
+#</p>
+#
+procedure jsf_validate(jfname, sfname, strict, error)
+   local json_date, json_schema
+
+   jerror := ErrorHandler(error,strict)
```
Don-Ward, line 1014: should `jerror` be `error`?

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+
+   json_data   := jtou(jfname , "f", strict, error)
+   json_schema := jtou(sfname , "f", strict, error)
+
+   return _jsvalidation(json_data, json_schema, json_schema)
+end
+
+procedure jss_validate(data, schema, strict, error)
```
Don-Ward, line 1022: Need to document the parameters. (I'm guessing that this is "string validate" as opposed to "file validate".  Also make the same fix to `jerror` as above (if needed).

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+#
+# <[param x is the table]>
+# <[param val is the name of the value of a property]>
+# 
+# <[returns "" if the valid otherwise returs the error string]>
+#
+#</p>
+
+procedure validate_property(x, val)
+   z := ""
```
Don-Ward, line 1045: Although it's optional, I think we should declare local variables (c, cr, i)

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+            return "expected object type, got: " || val
+            }
+         req := x["required"]
+         if \req then {
+            i := 1
+            cr := 0
+            while i <= *req do {
+               c := 0
+               every prop := key(val) do{
+                  if prop ~== req[i] then c := c + 1
```
Don-Ward, line 1058: Is there a reason for not using augmented operators? (this comment applies to several places in this procedure)

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+                  if prop == req[i] then cr := cr + 1
+                  }
+               if cr = *req then break
+                  if c =  *val then  z := "required properies missing"
+               i := i + 1
+               }
+            return z
+            }
+         maxProp := x["maxProperties"]
+         if \maxProp then write(space, "    maxProperties: ", left(maxProp, 20))
```
Don-Ward, line 1068: Where is `space` defined?

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+         }
+      "array": {
+         if type(val) ~== "list" then {
+            return "expected array type, got: " || image(val)
+            }
+         y := x["items"]
+         addItems := x["additionalItems"]
+         if \addItems then {
+            write(space, "   additionalItems: ")
+            dumpschema(x[additems])
```
Don-Ward, line 1081: where is `dumpschema` defined?

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+            enum := x["enum"]
+            if \enum then{
+               i := 0
+               c := 0
+               while i < *enum do {
+                  if val ~== enum[i] then c := c + 1
+                  i := i + 1
+                  }
+               if c ~= 0 then {
+                  return val || " is out of range, allowed values are: " || fullimage(enum)
```
Don-Ward, line 1102: `fullimage()` has not changed since its introduction to the repo in 2001.  Should we be using `ximage` instead?
This comment applies to several places in this procedure.

uni/lib/json.icn
```diff
@@ -991,3 +998,233 @@ end
...
+# <[param schRoot name of the schema file once more]>
+# The procedure _jsvalidation recusively goes into the object file and the corresponding level of the Schema file.
+# In each level the procedure  gets the value of each propertiy in the object file and calls the  validate_property() procedure.
+# with the value and the corresponding schema property.
+#
+# <[returns "" when valid, or the error string]>
+#
+#</p>
+
+procedure _jsvalidation(obj, sch, schRoot)
```
Don-Ward, line 1190: It's probably too large a change ... but
This procedure only returns the first error it finds, leading to many cycles of "find first error" then "fix first error".

Would it be possible to suspend, rather than return and, in the calling procedures, generate a list of errors rather than just a single error?

uni/progs/jsvalid.icn
```diff
@@ -0,0 +1,62 @@
...
+    write("\t -f : json filename")
+    write("\t -s : json schema file, default to <fnamet>.schema.json")
+    write("\t -h : show this help\n")
+    exit(-1)
+end
+
+procedure main(args)
+   local opts, jfname, sfname, ext
+   # : means that an input is specified
+   opts := options(args, "hs:f:")
```
Don-Ward, line 39: The error messages from `options()` do not include any clues about which options are available. I think it would be better to include a custom procedure that, at least, reminds the caller of the -h option to get help.

Don-Ward (changes requested):
I think this needs further work.

cjeffery (commented):
I concur with Don's comments.  Perhaps the comment requesting better test coverage can be deferred. Most of the other suggestions are easy to fix.

uni/lib/json.icn
Jafaral: old doesn't mean less better! ;). `ximage()` is good for debugging, whereas `fullimage()` is good for user output. Example:

ximage:
```
L199 := list(3)
   L199[1] := "Alice"
   L199[2] := "Rayya"
   L199[3] := "Rayya"
```
fullimage:
```
["Alice","Rayya","Rayya"]
```

uni/lib/json.icn
Jafaral: dropped

uni/lib/json.icn
Jafaral: dropped

tests/lib/jsontests.icn
Jafaral: of course!

Jafaral:
I addressed most comments.  I still need to fix the options. 

I did add more in-code comments, cleaned up, fixed, and expanded some of the validation cases.  I did add a few more tests for better coverage l. There are a lot more things that can be done, but this now provides good basic json validation.

uni/lib/json.icn
Don-Ward: Fair enough. In that case perhaps we should bring `fullimage` up to date: there are several things (e.g. patterns) that it doesn't deal with.

Jafaral: add it to out growing TODO list! :)

uni/lib/json.icn
Jafaral: Yes, that makes sense. I think I will defer this for now though.

uni/progs/jsvalid.icn
Jafaral: Not sure I follow. we do have `usage()` and it is called on various errors and the user can still to `-h` too.

Jafaral:
> I addressed most comments. I still need to fix the options.

I reviewed `options` not sure what is missing there.

uni/progs/jsvalid.icn
Don-Ward: The point I'm making is that without a third argument to `options() `if, for example, the user types `jsvalid -help` they will get the standard error message `Unrecognized option: -e` with no clue as to what the valid options are. They have to "know" that it's `-h` or no parameters at all to print the help.  See the `usage() `proc in `uni/progs/idxgen.icn` for an example of what I'm proposing (although it perhaps doesn't need to be as elaborate as that; something that listed a summary of the options after the error message would be fine).
If you don't want to do that, that's fine too but (in my view) it makes it slightly harder to use.

Jafaral: ah, well that is  easy, nothing but the best for you @Don-Ward  :)

PR #295, opened 2022-11-23
Update Makefile 

McChuck05:
The Makefile in /uni/uflex/test had two errors near the bottom that killed the make process.  I edited them, and left comments with the original line just above the edits.

### DIFF
--- a/uni/uflex/test/Makefile
+++ b/uni/uflex/test/Makefile
@@ -37,13 +37,15 @@ nummain.u: nummain.icn
 timetst2.u: timetst2.icn
 	$(UNICON) $(UFLAGS) -c timetst2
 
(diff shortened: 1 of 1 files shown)


Jafaral:
There is nothing wrong with the Makefile the way it is. the j0 and nnws targets are needed because those are the executable names. What platform are using?

Jafaral:
The CI test runs on Ubuntu/Centos/Debian/MacOS/Windows, and they all work fine without the change. Having multiple targets is supported by `make`, and also  you'd expect the  target to be the name of the executable. So even if we want to drop a target , we should drop the one with the post fix `.tst`.

Jafaral:
Unable to reproduce this issue.

PR #306, opened 2022-12-24
runtime: initialize a window pbuffer only if the window is GL2D

Jafaral:
PBuffer seems to be deprecated. glXCreatePbuffer fails on MacOS. So the GL 2D implementation still needs a fix, but this patch allows 3D graphics to work again at least.

### DIFF
--- a/src/runtime/rxwin.ri
+++ b/src/runtime/rxwin.ri
@@ -1590,18 +1590,20 @@ int wmap(wbp w)
 
    /*
     * Get PBuffer for offscreen rendering (OpenGL rendering only)
+    * TODO: migrate to use  framebuffer objects instead,
+    * PBuffers are deprecated in OpenGL 3.0 and not available on MacOS
     */
-   {
-   int pbufattrs[] = {GLX_PBUFFER_WIDTH, DisplayWidth(stddpy, wd->screen),
-                      GLX_PBUFFER_HEIGHT, DisplayHeight(stddpy, wd->screen),
-                      GLX_PRESERVED_CONTENTS, True,
-                      None};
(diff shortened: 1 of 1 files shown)


src/runtime/rxwin.ri
```diff
@@ -1591,17 +1591,17 @@ int wmap(wbp w)
...
-                      GLX_PBUFFER_HEIGHT, DisplayHeight(stddpy, wd->screen),
-                      GLX_PRESERVED_CONTENTS, True,
-                      None};
-   ws->pbuf = glXCreatePbuffer(stddpy, wd->configs[0],pbufattrs);
-   if (ws->pbuf == (GLXPixmap)NULL) {
-      /* consider freeing window binding/resources */
-      set_errortext(144);
-      return Failed;
-      }
+   if (wc->rendermode == UGL2D) {
```
Don-Ward, line 1594: Suggest adding a comment along the lines of 
> ToDo: Consider moving from Pixel Buffers to Frame Buffer Objects (FBOs)
>       as suggested by https://github.com/XQuartz/XQuartz/issues/264
>       which discusses the problem.
> 
If you want to change/improve the words (for example to mention explicitly that Apple have deprecated Pixel Buffers) then feel free.

Don-Ward (approved):
Looks OK (as far as I can see).  If you disagree with my suggestion to comment the change than that's fine.

src/runtime/rxwin.ri
Jafaral: Thanks @Don-Ward . That is the  issue I looked at when I did my fix. I will add a comment.

Jafaral:
@Don-Ward , I added a TODO in the code and also added comments to the commit message. Please merge if you are OK with that.

PR #308, opened 2023-01-18
Suppress writing to uniclass files whist importing for unicon -C

Don-Ward:


cjeffery (approved):
Looks pretty OK to me.

PR #322, opened 2023-02-13
Msys2 build

Don-Ward:


README.md
```diff
@@ -114,6 +114,50 @@ explicitly set the compiler as follows:
...
+```
+pacman -S mingw-w64-ucrt-x86_64-gcc
+pacman -S mingw-w64-ucrt-x86_64-diffutils
+pacman -S make
+pacman -S git
+```
+- Install the optional libraries for a full build (Unicon will build without them but some features
+will be absent).
+```
+pacman -S mingw-w64-x86_64-libjpeg-turbo mingw-w64-x86_64-libpng mingw-w64-x86_64-openssl
```
Jafaral, line 139: We need the ucrt version of these libraries:
```
pacman -S mingw-w64-ucrt-x86_64-openssl  mingw-w64-ucrt-x86_64-libpng 
mingw-w64-ucrt-x86_64-libjpeg-turbo
```

README.md
```diff
@@ -114,6 +114,50 @@ explicitly set the compiler as follows:
...
+- Download and run the installer from https://www.msys2.org/. At the time of writing it is called
+  `msys2-x86_64-20230127.exe` but it may be updated from that version.
+
+-  Go through the installation process to get a UCRT64 environment.
+-  Using the UCRT64 shell, Install tools required for the build:
+```
+pacman -S mingw-w64-ucrt-x86_64-gcc
+pacman -S mingw-w64-ucrt-x86_64-diffutils
+pacman -S make
+pacman -S git
```
Jafaral, line 134: just to be a bit concise 
```
pacman -S mingw-w64-ucrt-x86_64-gcc  mingw-w64-ucrt-x86_64-diffutils
pacman -S make git
```

Makefile
```diff
@@ -204,9 +210,9 @@ UIPL=$(ULROT)/ipl
 UPLUGINS=$(ULROT)/plugins/lib
 INST=$(SHTOOL) install -c
 F=*.{u,icn}
-Tbins=unicon icont iconx iconc unicont uniconx uniconc udb uprof unidep unidoc \
-	ui ivib patchstr iyacc rt.a rt.h
-
+Tbins=unicon$(EXE) icont$(EXE) iconx$(EXE) iconc$(EXE) unicont$(EXE) uniconx$(EXE) \
+      uniconc$(EXE) udb$(EXE) uprof$(EXE) unidep$(EXE) unidoc$(EXE) ui$(EXE) ivib$(EXE) \
+      patchstr$(EXE) iyacc$(EXE) rt.a rt.h
```
Jafaral, line 215: While we are at it, can you please fix a few of those names to their parametrized one. see line 103 to 109 above.  Also, on Windows these are missing the windows-specific binaries such as wiconx.exe and wicont.exe, but the variables 103 to 109 should cover those too if you add them all here.

Jafaral (changes requested):

Jafaral:
Another unrelated  fix but relevant for windows, can you  `chmod +w  src/gdbm/gdbm.h` and commit that as part of this PR.  for some reason that file is checked in with read-only permission, when we copy it over to rt/include and then do `make Pure` we get an error because it is read-only.

Makefile
```diff
@@ -28,8 +28,14 @@ default_target: allsrc
 	bin/unicon -features >> unicon-features.log
 	@echo ======================================== >> unicon-features.log
 	@cat unicon-features.log
+ifeq ($(MSYSTEM),UCRT64)
+	@echo "To finish, do either \"make WinPatch\" or \"make install\""
+	@echo "   make WinPatch completes the installation in situ."
```
Jafaral, line 33: Why is `make WinPatch` needed?  Binaries seems to already have the correct path already, here is what I got for example without having to patch:

```
Binaries at C:\msys64\home\WDAGUtilityAccount\unicon\bin\
```

Makefile
Jafaral: you cab ignore this one. Just realized we have written this so we can handle iconx/uniconx renaming already. One thing to add if you want to make the `make install`  "complete" on windows, is to add window-specific binaries such as `wiconx` `wicont`, `wunicon`, etc, installed too.

Jafaral: Notice that the you don't have to make that a windows conditional install. The install rule is smart enough to skip the ones that don't exist.

Jafaral (approved):

PR #326, opened 2023-02-24
Add the Bitman plugin to Appendix C of the Unicon book.

Don-Ward:
Add a short description of plugins (which could perhaps be extended) plus a description of the Bitman plugin to Appendix C.


cjeffery (commented):
Glad to see some documentation of Plugins; would like to see more on their general architecture (probably elsewhere in the book, not here). Looks good to me. Jafar should be the approver since Plugins are his thing.

Jafaral (approved):

PR #329, opened 2023-03-04
Add the SQLite plugin to the Unicon book.

Don-Ward:


cjeffery (commented):
I am thrilled that Unicon programmers will be able to use SQLite. You give a pretty good accounting of it being a potentially faster and/or in-memory local database. We kind of need to provide a better explanation of where this sits in the space between GDBM and ODBC, and what SQL it can do and what parts of SQL it can't do. An example that migrates a SQLite database to an ODBC database or vice versa would be interesting.

Aside from minor stylistic matters such as avoiding single-sentence paragraphs and not starting sentences with a code statement, the main issue I have with this proposed addition is that it creates two different SQL API's where programmers would be better served having one unified SQL API to the extent that that is possible.  I don't know whether to recommend approving, or to recommend that the ODBC API be brought into compliance with the SQLite API, or to recommend that the SQLite API be brought into compliance with the ODBC API, or some intermediate hybrid be created that allows common coding for both.

It would be reasonable to construct an argument that it is impossible to unify the API's for some specific technical reason to be stated. It would be reasonable to present this as providing a higher level OO API as a separate thing on top of one or more lower-level built-in API's.  But at the moment we are saying "you have to learn two different ways to use SQL" without giving a rationale for this questionable language design decision.

PR #340, opened 2023-05-03
Param names in comments are now italicized + some small cleanup

Mj10111:
This is a small (and hopefully final) push, so there shouldn't be many issues, if there are any.


Jafaral:
Can  you please use a better title for this PR ? what do you mean by "Comment" ? What does this PR do or add ?

Also, the second commit says is cleaning up white spaces,  but it looks to me as it adds white space,. not remove it.

Jafaral:
included in #345 , closing ...

PR #343, opened 2023-05-21
add $$= syntax

cjeffery:
Tweak iyacc to interpret "$$=" as "$$:=" under Icon/Unicon, allowing the same .y file to work for both Unicon and Java, for example.

### DIFF
--- a/src/iyacc/reader.c
+++ b/src/iyacc/reader.c
@@ -76,6 +76,14 @@ void get_line(void)
     {
 	line[i]  =  c;
 	if (c == '\n') { cptr = line; return; }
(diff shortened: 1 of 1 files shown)


cjeffery:
Apparently in Github we are unable to say Dollarsign Dollarsign Equals as "$$=", but that is what this is for; slight multi-language semantic action support.

Jafaral:
you can use the  tick "`"  character to format inline code. for multiline code format you can have a line with 3 ticks before and the same after.

example:  `$$=`

or 
```
$$=
$$:=
```

PR #355, opened 2024-01-06
runtime: handle some SSL errors

Jafaral:


Don-Ward (approved):
After applying these these changes to a test program that was sometimes crashing after returning an illegal string descriptor (usually within 5 minutes running) no crashes occurred during twenty four hours. So it looks like the problem has been fixed.

PR #356, opened 2024-01-24
unidoc: close the file when done with it

Jafaral:


Makefile
```diff
@@ -61,7 +61,6 @@ help:
 	@echo "        sh configure --build=x86_64-w64-mingw32"
```
Don-Ward, line 61: This will work on the MSVCRT option documented in the README. I'm not sure it will work on the UCRT64 option

Don-Ward (commented):
The help in the Makefile may need some more work. Apart from that, it looks good.

Makefile
Jafaral: I didn't change that line. The only reason I included the change below is that the makefile was out of sync with the Makefile.in. I.e, the diff is automatically generated. So, I just fixed that.

PR #357, opened 2024-01-25
ulsp: Add dot and dual colon completion support

mstreeter10:


uni/ulsp/workspace.icn
```diff
@@ -627,17 +638,54 @@ class Context(uniAll, uri, line, lineNum, charNum, objectName, methodName, conte
...
+      }
+
+      #-----------------------------------------------#
+      #--------- Case "package dual colon" -----------#
+      #-----------------------------------------------#
+
+      line ? {
+         tab(charNum)
+         move(-2)
+         if move() == ":" & move() == ":" then {
```
Jafaral, line 678: this is the same as 
```
if  = "::" then {
```

Jafaral: I guess it is a bit different. Since `=` will only move the scan position if there is a match.

Jafaral (approved):

PR #358, opened 2024-02-10
runtime: Make "mono" select a monospaced font.

Don-Ward:
Previously, "mono" and "fixed" were treated as the same thing and asked for Lucida Sans, which is a proportionally spaced font. Now, "mono" asks for Lucida Console, which is monospaced, and "fixed" selects Lucida Sans (with the correct flags).

### DIFF
--- a/src/runtime/rmswin.ri
+++ b/src/runtime/rmswin.ri
@@ -1674,9 +1674,13 @@ HFONT mkfont(char *s, char is_3D)
        * This is a legal Icon font spec.
        * Check first for special "standard" family names.
        */
-      if (!strcmp(family, "mono") || !strcmp(family, "fixed")) {
-	 stdfam = "Lucida Sans"; /* Lucida Console? */
+      if (!strcmp(family, "mono")) {
+	 stdfam = "Lucida Console";
 	 flags |= FONTFLAG_MONO + FONTFLAG_SANS;
+	 }
+      else if ( !strcmp(family, "fixed")) {
+	 stdfam = "Lucida Sans";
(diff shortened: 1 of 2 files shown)


src/runtime/rmswin.ri
```diff
@@ -1674,10 +1674,14 @@ HFONT mkfont(char *s, char is_3D)
        * This is a legal Icon font spec.
        * Check first for special "standard" family names.
        */
-      if (!strcmp(family, "mono") || !strcmp(family, "fixed")) {
-	 stdfam = "Lucida Sans"; /* Lucida Console? */
+     if (!strcmp(family, "mono")) {
```
Jafaral, line 1677: indentation is off

src/runtime/rxwin.ri
```diff
@@ -5698,10 +5698,14 @@ void mkfont(char *s, char is_3D)
        * This is a legal Icon font spec.
        * Check first for special "standard" family names.
        */
-      if (!strcmp(family, "mono") || !strcmp(family, "fixed")) {
-	 stdfam = "Lucida Sans"; /* Lucida Console? */
+     if (!strcmp(family, "mono")) {
```
Jafaral, line 5701: same here

src/runtime/rmswin.ri
Don-Ward: That willl be the tabs (sigh).  OK, I'll fix it.  There are are times (most of the time, actually) when I wish we could replace **every** tab in the source code with spaces.

Jafaral (approved):

PR #360, opened 2024-02-21
update the config script to latest

Jafaral:


Don-Ward (commented):
These are really two separate changes. I wonder if it would be better to keep them separate, just in case we end up reverting config.guess.  Would it be worth keeping a copy of the old config.guess for a time in case somebody has a problem with the new one? In that case, it would be easy (for them) to work around the problem. I'm not sure what "the time" should be: maybe a few months (or until after the next release).

PR #362, opened 2024-02-22
runtime: don't exit the main thread if a coexp attempted to gc it

Jafaral:


src/runtime/rmemmgt.r
```diff
@@ -1535,6 +1535,8 @@ static void cofree()
 
          xep = *ep;
          *ep = (*ep)->nextstk;
+         if (xep->program == &rootpstate || xep->id == 1)
```
Don-Ward, line 1538: This causes build errors for some configurations.

Don-Ward (changes requested):
I think the headline comment is misleading. It says we should never GC the main thread, which is obviously wrong.  I _think_ the mod is attempting to stop the recovery of the activation block for the main thread (which makes sense, because nobody activated it). 

However, the big problem is that the ci fails. I put the patch into my (up to date) master branch on my local machine and it failed there too.

src/runtime/rcoexpr.r
```diff
@@ -577,19 +577,25 @@ void coclean(struct b_coexpr *cp) {
...
+         * If the thread is cleaning itself, exit, what about tls chain?
+         */
+         cp->alive = -1;         /* signal thread to exit */
+         cp->have_thread = 0;
+   #ifndef NO_COEXPR_SEMAPHORE_FIX
+         if (cp->semp) {SEM_CLOSE(cp->semp); cp->semp = NULL;}
+   #endif                  /* NO_COEXPR_SEMAPHORE_FIX */
+         pthread_exit(0);
+      } else if (cp->id == 1 && cp->id == curtstate->c->id) {
+         /* FIXME: Main thread, should just return? */
```
Don-Ward, line 593: Should this FIXME be here?  If yes, it's new code; why aren't we fixing it now?

Don-Ward (commented):
Looks OK. I'm not sure if the FIXME describes what has been done (and should be removed) or if it is questioning the new code .

src/runtime/rcoexpr.r
Jafaral: Left it there intentionally because I wasn't there if this has any side effects.

PR #368, opened 2024-03-03
tests: fix a few posix issues. Skip test if feature isn't supported

Jafaral:
If the test standard output contains the text "This program requires", the test is considered skipped and the output is reported as is. Ex:
 
```
[Testing utf8]... OK
[Testing utf8_ovld]... This program requires operator overloading.
[Testing var]... OK
```


Don-Ward:
This PR could be your opportunity to alter the `tests/unicon/makefile `introduced in PR#363 to how you'd like it to be.

PR #370, opened 2024-03-04
Corrections to the fixed point arithmetic class.

Don-Ward:
Correct the div method to give control over the number of places. Fix the div method to give the correct rounding.
Add tests for the above.
Add the type names that are used by iconc.

### DIFF
--- a/uni/lib/FxPt.icn
+++ b/uni/lib/FxPt.icn
@@ -83,9 +83,10 @@ class FxPt: _FxPt()
    # Divide another fixed-point value or number from this FxPt.
    # <[returns result of the division as a FxPt]>
    #</p>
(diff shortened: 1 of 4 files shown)


uni/lib/FxPt.icn
```diff
@@ -329,6 +332,8 @@ HIDE # --------Overloaded Infix operators and internal methods----------
          "integer": { return p1 }
          "real"   : { return integer(p1)}
          "string" : { return \integer(p1)}
+         "FxPt__mdw_inst_mdw" | "_FxPt__mdw_inst_mdw" |
```
Jafaral, line 335: should we change these fields name first?

StephenWampler (approved):
Looks fine to me.

PR #373, opened 2024-03-04
iconc: rename the instance state field

Jafaral:


src/iconc/cmem.c
```diff
@@ -103,7 +103,7 @@ rec_is_obj(name)
 
    if ((len = strlen(name)) < 15)
       return 0;
-   if (strcmp("__mdw_inst_mdw", (char *)(name + len - 14)) == 0)
+   if (strcmp("__state", (char *)(name + len - 14)) == 0)
```
Don-Ward, line 106: I think we need to change the 14 to 7

src/iconc/ica.c
```diff
@@ -342,7 +342,7 @@ find_clsinst_call(clsname)
    struct pcall * pcall;
 
    clslen = strlen(clsname);
-   clslen -= 14/* "__mdw_inst_mdw" */;
+   clslen -= 14/* "__state" */;
```
Don-Ward, line 345: 14 --> 7

Don-Ward (changes requested):
Need a couple of  changes (replace "14" by "7") to reflect the difference in string lengths.

PR #380, opened 2024-03-08
Remove MacGraph configuration

Don-Ward:


Jafaral (changes requested):
mixed in with space cleanup

PR #388, opened 2024-03-14
Remove old-style C function declarations.

Don-Ward:
Change old style function declarations to the new style. Modify rtt to emit new style function declarations.


Jafaral (approved):

src/rtt/rttinlin.c
```diff
@@ -1567,6 +1552,183 @@ struct node *n;
    fname = cname;
    line = 7;
 
+#ifndef OLD_SCHOOL
```
Jafaral, line 1555: I suggest we use `LEGACY_FUNC_PROTOTYPE` , just to make it clearer, instead of `OLD_SCHOOL`.

Jafaral (approved):
I ran the tests and inspected the results with and without this PR, everything looks good.

PR #391, opened 2024-03-25
runtime: remove some build warnings

Jafaral:


src/runtime/rxwin.ri
```diff
@@ -4060,7 +4060,8 @@ int strimage(wbp w,
       }
    }
    else {
-      int lastr=-1, lastg=-1, lastb=-1, count = 0;
+      int lastr=-1, lastg=-1, lastb=-1;
+      //int count = 0;
```
Don-Ward, line 4064: I've been looking at the history of this block of code. `count` was introduced by commit f1c12578 in 2007. It wasn't used then (except to be incremented for no apparent purpose) and it has never been used up to today. I'm content for you to comment it out, but I wonder why we don't just delete it.

Don-Ward (approved):

PR #394, opened 2024-03-25
Complex class: Workaround Unicon -C problems.

Don-Ward:
The present optimising compiler has problems with the complex class. The workaround(s) are to remove the Object superclass and to provide an initially section to set the default values.
These changes can be reverted when the problems are solved.

### DIFF
--- a/uni/lib/complex.icn
+++ b/uni/lib/complex.icn
@@ -7,8 +7,6 @@
 
 package math
 
-import lang
-
 #<p>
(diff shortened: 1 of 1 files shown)


StephenWampler (commented):
I'm okay with this, but was wondering if the compiler had problems with the use of the superclass, the default value settings in the class member list "(r:0,i:0)", or both?

Jafaral (approved):

PR #399, opened 2024-03-27
IPL: suppress clang's deprecated-non-prototype warning.

Don-Ward:
It was decided to retain the old style function declarations in the IPL. So we suppress the clang warnings about them.


Jafaral:
Let us see if #400 actually clears the warnings rather than suppress them.

PR #401, opened 2024-03-29
ci: enable werror on a few platforms, remove skip test

Jafaral:


Don-Ward (approved):
Looks even better than it did three days ago.

PR #415, opened 2024-04-04
udb: fixed safe_filelist

austin-barr:


uni/udb/symtab.icn
```diff
@@ -522,7 +522,8 @@ method lookup_fileline(filename, lineno)
    local symt, temp
 
    if \filename & \lineno then {
-      if not member(safe_filelist, filename) then fail
+      every f := !safe_filelist
```
Jafaral, line 525: This seems to be a good use case for `set()` since we only care about membership, and it is much more efficient than looping over all of the list elements. 
```
safe_filelist := set()
```
`member()` can then be used with `safe_filelist`

Jafaral:
Can you squash the two commits together?

Jafaral (approved):

PR #417, opened 2024-04-09
runtime: Fix bugs in string to int conversion when the str has a suffix

Don-Ward:
The previous code had two problems:
(1) Overflows were sometimes not detected when the string had a suffix. (2) Strings with a suffix that were big integers would fail (strings
   with suffixes that were not a big integer worked).

This commit also fixes a related bug in numeric(), which uses string conversion.

### DIFF
--- a/src/runtime/cnv.r
+++ b/src/runtime/cnv.r
@@ -817,7 +817,7 @@ static int tmp_str(char *sbuf, dptr s, dptr d)
  * dp_pnmcmp - do a string comparison of a descriptor to the procedure
  *   name in a pstrnm struct; used in call to qsearch().
  */
-int 
+int
 dp_pnmcmp (struct pstrnm *pne, struct descrip *dp)
 {
    struct descrip d;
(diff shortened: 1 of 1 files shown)


src/runtime/cnv.r
```diff
@@ -1114,11 +1121,65 @@ static int ston(dptr sptr, union numeric *result)
...
+            * 32 bit m/c  [KMG] are  small integers, [TP] are large integers
+            * 64 bit m/c  [KMGTP] are all small integers
+            */
+           switch (*esave) {
+           default: return CvtFail; /* unkown suffix */
+           case 'k': case 'K': { MakeInt(1024L,&db); break; }
+           case 'm': case 'M': { MakeInt(1024L*1024,&db); break; }
+           case 'g': case 'G': { MakeInt(1024L*1024*1024,&db); break; }
+#if WordBits == 64
+           case 't': case 'T': { MakeInt(1024L*1024*1024*1024,&db); break; }
```
Jafaral, line 1156: `L` should be dropped from all cases because it forces a `long int` evaluation which is not the same on all platforms.

Jafaral (approved):

PR #420, opened 2024-04-12
Fix build errors on Windows when using --disable-graphics

Don-Ward:
The fix consists mostly of moving routines that are not graphics oriented to places that included them even if Graphics is not defined.

### DIFF
--- a/src/runtime/fsys.r
+++ b/src/runtime/fsys.r
@@ -2054,6 +2054,7 @@ function{0,1} seek(f,o)
 end
 
 #ifdef PosixFns
+
 "system() - create a new process, optionally mapping its stdin/stdout/stderr."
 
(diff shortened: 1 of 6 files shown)


src/runtime/fsys.r
```diff
@@ -2054,6 +2054,63 @@ function{0,1} seek(f,o)
 end
 
 #ifdef PosixFns
+#ifdef MSWindows
+int mswinsystem(char *s)
```
Jafaral, line 2058: Should probably move this to `rsys.r`. In general the `f` version of the file for the language visible functions, and the `r` file is for runtime support of those functions. eg. `fstruct.r -> `rstruct.r`,  `fsys.r` -> `rsys.r`.

Don-Ward: OK.

PR #425, opened 2024-04-16
ulsp: add go-to definition and format whitespace

mstreeter10:


uni/ulsp/definition.icn
```diff
@@ -0,0 +1,62 @@
...
+         }
+         i := i + 1
+      }
+
+      close(source)
+
+      \selectedLine ? {
+         tab(many("\ \t"))
+         tab(many(wchar))
+         tab(many("\ \t"))
```
Jafaral, line 51: `many()` takes a `cset`, so it it is "slightly" faster if you do `tab(many('\ \t'))` because you avoid `string` to a `cset` conversion. This is applicable to several string scanning functions. `find()` is an exception, it take a string.

Jafaral (changes requested):

Jafaral (commented):
A couple more minor comments.

uni/ulsp/definition.icn
```diff
@@ -0,0 +1,63 @@
...
+
+      if source := open(item["source"]) then {
+         every line := !source do {
+            if ReMatch("^[\ \t]*" || item["type"] || "[\ \t]+" || item["name"], line) then {
+               selectedLine := line
+               break
+            }
+            i := i + 1
+         }
+         close(source)
```
Jafaral, line 47: `close(source)` need to go before line 46 to ensure source did in fact `open()`

Jafaral (approved):

PR #427, opened 2024-04-18
udb: eval no longer skips first variable in expression

austin-barr:


uni/udb/exparser.icn
```diff
@@ -23,6 +23,13 @@ procedure getIdentifiers(expr)
 		if i === IDENT then
          found := yytext
 	}
+   if \found then {
+      if i === LPAREN then
+         insert(procs, found)
+      else
+         insert(vars, found)
+      found := &null
+   }
```
Jafaral, line 32: The indentation of the whole function looks odds due to a mix of tab and spaces. Can you please fix it? otherwise the change looks fine.

PR #429, opened 2024-04-19
runtime: fix buffer overflow with wildcard shell expansion in open()

Jafaral:


Don-Ward (changes requested):
Although this change will likely fix most problems encountered in practice it isn't bomb proof: the string is included twice in the output buffer and there are extra characters. We could replace the 1st condition with  `((2*strlen(fnamestr)+58) < MaxPath)` but that's fragile because it depends on the details of the command string on line 1112. Instead, I think we should replace `sprintf` on line 1112 with `snprintf` and check for failure because of overflow.
It wouldn't hurt to replace line 1104 with `(2*strlen(fnamestr) < MaxPath`. That will catch most cases and we'll only fail the `snprintf` test for the edge case where `strlen(fnamstr) `is just under half the length of MaxPath, but the extra characters make it overflow.

In passing, I'm not sure why the nearby comments on line 1108 and 1118 are in the past tense ("attempted to .."). I think "Attempt to .." more accurately describes what is going on.

Jafaral:
I think the condition should still be `strlen(fnamestr) < MaxPath`, since `fnamestr` is assumed to be a path, that is why I added that test. I did allocate the `3 * MaxPath` buffer to account for `2*strlen(fnamestr)` plus  some extra room for the shell script. For any `MaxPath>58` the buffer will not overflow, but I can but explicit there and be exact.

Issue #422, opened 2024-04-15
invoke method is not working


IanTrudel:
The package lang provides the utility method invoke but it does not work. It was confirmed when I ran objecttests. Unicon version 13.3 on Linux.

Output from unicon/tests/lib/objecttests.icn:

```
Check to see if d is an instance of Class.

d is NOT an instance of Class!

Test string invocation of method names...should see '123'


List all the methods in class D...

DTest_write

Check the inheritance hierarchy

ATest is NOT an instance of Class
ATest is NOT an instance of BTest
BTest is NOT an instance of ATest
BTest is NOT an instance of BTest
CTest is NOT an instance of Class
CTest is NOT an instance of ATest
DTest is NOT an instance of Class
DTest is NOT an instance of ATest
DTest is NOT an instance of BTest
DTest is NOT an instance of CTest
ETest is NOT an instance of Class
ETest is NOT an instance of FTest
FTest is NOT an instance of ETest

d inherits from: DTest ATest CTest lang::Object

e inherits from: ETest FTest lang::Object ETest

g fields: one two three
h fields: four five six one two three
h.one -> 1
h.one -> One
Attempt to set non-existent field: ok - field doesn't exist
```

StephenWampler:
Okay, there are lots of interesting things going wrong here.  Starting with the simple and heading towards the complex:

1) The class "Object" no longer subclasses the class "Class", so objecttests.icn needs to change all references to the Class class into references to the Object class.

2) The procedure instanceof in uni/lib/predicat.icn currently calls the istype procedure.  istype appears to be broken, possibly in several ways (for one thing it calls get_type but get_type doesn't traverse the inheritance hierarchy).  A workaround would be to modify the instanceof method to call lang::Type instead of istype.  The correct thing may be to fix get_type but who knows what that may break?

3) In uni/lib/object.icn, the method invoke calls hasMethod, which has the line:

    return mName == genMethods()

[...]

PR #432, opened 2024-04-20
Introducing Unit Test Framework

IanTrudel:
- Provides a TestCase, TestSuite, TestReporter and a Mock class.
- Testing is bootstrapped using the new unittest package to test the said package.

### DIFF
--- a/uni/lib/unittest.icn
+++ b/uni/lib/unittest.icn
@@ -205,7 +205,7 @@ class Mock(__expectations)
    end
 
    method invoke(mname, args)
-      self.verify(mname, args)
+      self.verify(mname, args) | fail
       return self.__expectations[mname]
    end
 
--- a/uni/lib/unittest.icn
+++ b/uni/lib/unittest.icn
@@ -104,7 +104,7 @@ end
 
 class TestSuite(__tests, __testReporter)
    $define EXIT_SUCCESS 0
-   $define EXIT_FAIL 1
+   $define EXIT_FAILURE 1
 
    method add(testCase)
       if (classname(testCase) ~=== "unittest__TestCase") then {
@@ -147,7 +147,7 @@ class TestSuite(__tests, __testReporter)
       self.__testReporter.runtime(((t1.sec - t0.sec) * 1000000 + (t1.usec - t0.usec)) / 1000000.0)
       summary := self.__testReporter.summary()
 
-      return if summary.failed > 0 then EXIT_FAIL else EXIT_SUCCESS
+      return if summary.failed > 0 then EXIT_FAILURE else EXIT_SUCCESS
    end
 
    method output()
--- a/uni/lib/unittest.icn
+++ b/uni/lib/unittest.icn
@@ -52,9 +52,31 @@ class TestCase : Object(__status)
       return
    end
 
+   method assertFail(expression)
+      if not @expression[1] then return
+
+      write("Assertion failed.")
+      self.__status := STATUS_FAIL
+      fail
+   end
+
    method assertEqual(expected, actual)
-      if expected ~== actual then {
+      if type(expected) ~== type(actual) then {
+         write("Assertion failed: expected value of type ", type(expected), ", but got ", type(actual))
+         self.__status := STATUS_FAIL
+         fail
+      } else if expected ~== actual then {
          write("Assertion failed: expected ", expected, ", but got ", actual)
+         self.__status := STATUS_FAIL
+         fail
+      }
+
+      return
+   end
+
+   method assertNotEqual(unexpected, actual)
+      if type(unexpected) === type(actual) & unexpected === actual then {
+         write("Assertion failed: unexpected ", unexpected, ", but got ", actual)
          self.__status := STATUS_FAIL
          fail
       }
--- a/uni/lib/unittest.icn
+++ b/uni/lib/unittest.icn
@@ -1,3 +1,17 @@
+############################################################################
+#
+#       File:     unittest.icn
+#
+#       Subject:  An object-oriented unit test framework based on Kent Beck's SUnit.
+#
+#       https://web.archive.org/web/20150315073817/http://www.xprogramming.com/testfram.htm
+#
+#       Author:   Ian Trudel
+#
+#       Date:     April 19, 2024
+#
+############################################################################
+
 # TODO: assertions
 #     - asserting objects
 #     - asserting nested values (list, table, objects, etc.)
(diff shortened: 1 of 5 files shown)


uni/lib/unittest.icn
```diff
@@ -0,0 +1,166 @@
...
+   end
+
+initially
+   /__status := STATUS_SUCCESS
+end
+
+class TestSuite(__tests, __testReporter)
+   method add(testCase)
+      cname := classname(testCase)
+      if (\cname & (not (cname == "unittest__TestCase"))) then {
```
Jafaral, line 72: slightly shorter? `if classname(testCase) ~=== "unittest__TestCase" then...`

uni/lib/unittest.icn
```diff
@@ -0,0 +1,166 @@
...
+         self.__status := STATUS_FAIL
+         write("Assertion failed.")
+         fail
+      }
+
+      return
+   end
+
+   method assertEqual(expected, actual)
+      if expected ~== actual then {
```
Jafaral, line 56: Is this supposed to test strings only?

Jafaral: Looking at use cases, this should be `expected ~=== actual ` I think.

uni/lib/unittest.icn
```diff
@@ -0,0 +1,166 @@
...
+               write(ximage(mname))
+               test.__reset()
+               test.setup()
+               mname(test)
+               test.teardown()
+
+               self.__testReporter.result(cname, mname, test.__status)
+               write()
+            }
+         }
```
StephenWampler, line 102: cname := classname(test)

         test.setupClass()
         every image(!test.__m) ? {
            mname := (="procedure ", tab(0)) | next
            if mname ? (tab(*cname+1), ="_", match("test")) then {
               write(ximage(mname))
               test.__reset()
               test.setup()
               mname(test)
               test.teardown()

               self.__testReporter.result(cname, mname, test.__status)
               write()
            }
         }

(1) Is the intent here to only check a 'locally'-defined method?  As written using the tab(*cname+1) only a locally defined method is found.  If there isn't one this won't find an inherited method of the same name.

(2) Shouldn't you break out of the inner every if the method is found?

uni/lib/unittest.icn
IanTrudel: - [ ] Update how to ensure the class added is not directly an instance of TestCase.

uni/lib/unittest.icn
IanTrudel: I tried ~= that required two integers but ~== did fine for now. It would convert the numbers before the conversion into strings anyway. Not sure what the advantage of ~=== would be? Anyone could give me an example?

uni/lib/unittest.icn
IanTrudel: 1) Any given test class is a subclass of TestCase. The setupClass,(), setup(), teardown() and teardownClass() are ran exactly where they are supposed to be ran. Subclassing TestCase actually allow you to do what you want with those methods, for example, I reset values in setup() to ensure each test start afresh. setupClass could connect you to a DB or open a file whereas  teardownClass would close them.

Subclassing a subclassed class of TestCase is a not typical scenario. Using multiple inheritance hasn't been taken into account for now. According to https://github.com/uniconproject/unicon/issues/422, we would have to sit and think what the right approach it should be, or even if we need this.

This is worth mentioning that the implementation is not mutex friendly just yet, as `tests/unicon/tester.icn` is. 

For the rest, any method starting with "test" are to be run and anything else is not. It allows the possibility to define utility methods that may be useful within some tests, called at the discretion of the programmer within the tests. 

My current approach is to provide a Unit Test Framework that is fairly conventional to other TDD frameworks. 

2) No, a test could have zero, one or many test methods. It has to search the whole __m space.

uni/lib/unittest.icn
Jafaral: `===` supports all types and nit type specific like `=` and `==`.  It works with &null, tables and lists for example. Notice that with structures the check is just for the reference and not actual contents. So
```
L1 := L2 := [ ]
L3 := [ ]
 ```
Ao in the above you will get `L1 === L2`, but `L1~===L3`

Comment (IanTrudel) [reply uni/lib/unittest.icn:56]: The problem I see is that `===` prevents data conversion. `1 == "1"` will be valid but `1 === "1"` will fail. Here `L1 == L2´ will cause a runtime error but `L1 === L2´ will succeed. This is the reason I came up with assert { expression } based on t.verify { }. I also noticed that an assertion with parameters that fail won't even enter the assertion method, hence assert and verify use {}. If you have a solution for all those problems, I'm listening!

Comment (Jafaral) [reply uni/lib/unittest.icn:56]: Why do you want 1 to equal "1"?  for a test framework I think you have to be a bit more explicit about type conversion. Because by that logic all of the following would  still fail:
```
 "1" == "1.0"
   1 == "1.0"
   1 == "01"
```
If I want to allow mixing types, I'd do something like:
```
expected ~= actual & expected ~== actual & expected ~=== actual
```
That would allow all possible type coercions to happen in the "correct" order, and cover the cases I listed above and also the structures too.

Comment (IanTrudel) [reply uni/lib/unittest.icn:56]: My initial idea was to make assertEqual as convenient as possible and fully benefiting from implicit type conversion. You have however gave me to ponder about whether this should be implemented or not. I further investigated your solution and it would crash if `expected` is of type string. The edge cases would have to be handled. Perhaps ensuring matching types is the programmer's responsibility.

Comment (Jafaral) [reply uni/lib/unittest.icn]: my bad. `=` and `==` would cause a runtime error if you supply the wrong time operand. To guard against that I'd try:
```
integer(expected) ~= integer(actual) & string(expected) ~== string(actual) & expected ~=== actual
```

Jafaral: And now that might be the wrong logic!

Jafaral:
It is better to rebase your branch on top of the master branch instead of merging.

IanTrudel:
> It is better to rebase your branch on top of the master branch instead of merging.

Will do next time. Thank you!

IanTrudel:
@Jafaral sorry about the merge. I did set git to rebase but something clearly went wrong.

IanTrudel:
@Jafaral I resent you an invite to my fork. Would you be so kind to arrange so this can be merged? Otherwise I can redo it using rebase. My next contribution is in the works but I will be careful this turn around. :))

Jafaral:
@IanTrudel , I pushed a new branch to your repo with rebase and cleanup of your work. See the new PR I opened from your repo. Feel free to checkout the new branch at your end and make changes as you see fit.

Jafaral:
As a summary, here is what I did:
- rebase
- squashed intermediate fixes into their respective commits.
- cleaned up some formatting.

Jafaral:
Continued in #507

PR #439, opened 2024-04-25
Fix instanceof and istype procedures

StephenWampler:
This modifies the istype procedure and, by inference, the instanceof procedure to correctly recognize superclasses.  This is also the first step in fixing the invoke method.


Jafaral (changes requested):
This history is mixed up  on this PR. We will need to fix that first.

Jafaral:
Looking at the history, it looks like you use the `master` branch for development. So when you merge new  merge updates from upstream Unicon, your local master branch ends up  diverging and creating issues.

StephenWampler:
Yeah, things got confused when I tried to repair a "Detached HEAD warning" (which, apparently isn't really a bad thing anyway...).  Suggestions on how to fix?  In the meantime, I can cancel this pull request as I've now also have a fix for the invoke() issue that could be included.  I wait on cancelling, though, until I learn how to fix the "master/HEAD" issue!

PR #452, opened 2024-05-02
Do not strip symbols from iconx.

Don-Ward:
Plugins require the symbols to be present in iconx to load successfully.

### DIFF
--- a/src/runtime/Makefile
+++ b/src/runtime/Makefile
@@ -57,21 +57,19 @@ OBJS=	$(XOBJS) $(EXEICONTARGET)
 interp_all: update_rev 
 	$(MAKE) $(UNICONX)
 
-# Some systems require global symbols to be left in iconx for loadfunc() to work
-# but some versions of strip don't have the -x flag (e.g. strict POSIX compliance),
-# so only use strip -x when it is known to be needed and available (MacOS).
-
 # NTCONDEP is set to NTCON on windows only to add a dependency that sets NTCONSOLE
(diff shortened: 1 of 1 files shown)


src/runtime/Makefile
```diff
@@ -57,21 +57,19 @@ OBJS=	$(XOBJS) $(EXEICONTARGET)
...
-# but some versions of strip don't have the -x flag (e.g. strict POSIX compliance),
-# so only use strip -x when it is known to be needed and available (MacOS).
-
 # NTCONDEP is set to NTCON on windows only to add a dependency that sets NTCONSOLE
 $(UNICONX) $(ICONXEXE): $(NTCONDEP) $(OBJS) ../../rt/lib/libucommon.a ../../rt/lib/libuconsole.a ../../rt/lib/libgdbm.a ../../rt/lib/libtp.a
 	$(CC) $(LDFLAGS) $(WSUBSYS) $(WSTKLDFLAG) $(RLINK) -o $(UNICONX)  $(OBJS) $(XLIBS) $(RLIBS) $(XL) $(LIBSTDCPP) $(WGLIBS) -luconsole
 	cp $(UNICONX)$(EXE) ../../bin
-	$(STRIP) ../../bin/$(UNICONX)$(EXE);
+# Symbols are needed for plugins to load successfully
+#	$(STRIP) ../../bin/$(UNICONX)$(EXE);
```
Jafaral, line 65: Should we leave that in, and use `strip -g` instead ?

Don-Ward: There is no "-g" option on macOS.

PR #453, opened 2024-05-02
ulsp: socket reading fix

mstreeter10:
Fixes unhandled case where client hasn't finished sending data for body of a header + body pair.

### DIFF
--- a/uni/ulsp/server.icn
+++ b/uni/ulsp/server.icn
@@ -39,7 +39,7 @@ class Server(
 
       repeat {
          every request_body := get_request(sock) do {
-            jsontable := jtou(request_body)
+            jsontable := jtou(request_body) | next
 
             request_method := jsontable["method"]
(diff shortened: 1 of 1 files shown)


uni/ulsp/server.icn
```diff
@@ -87,9 +87,13 @@ class Server(
                if pos(0) then {
                   request_body := ready(sock, len)
                }
-               else
+               else {
                   request_body := move(len)
-                  msg := tab(0)
+               }
+               until *request_body = len do {
```
Jafaral, line 93: This had the potential of running infinitely. I recommend adding a `select()` with a 5 seconds   timeout, maybe in a loop with 1 second increment, after that, consider that a malformed msg and move on.

Jafaral (changes requested):

Jafaral (approved):

PR #454, opened 2024-05-02
udb: socket reading fix

mstreeter10:
Fixes unhandled case where client hasn't finished sending data for body of a header + body pair.


Jafaral (changes requested):
Same  comment from ulps pr.

Jafaral (approved):

PR #458, opened 2024-05-04
config: keep global symbols on macos for dynamic loading

Jafaral:


Jafaral:
@Don-Ward  does this solve the issue at #452 ?

PR #476, opened 2024-09-15
ulsp: unidoc chokes on 'parser', skip it for now

Jafaral:


uni/ulsp/database.icn
```diff
@@ -96,14 +96,15 @@ class LSPDB(
...
-      local dirPath, db_paths
+      local dirPath, db_paths := [], dir
 
       dirPath := unicon_dir
 
-      db_paths := ipaths_get()
-
-      pop(db_paths) # remove the current directory
-
+      write("Looking in:")
```
Don-Ward, line 103: Should there be some extra writes that print out each directory that is added?

uni/ulsp/database.icn
```diff
@@ -96,14 +96,15 @@ class LSPDB(
...
+      every dir := !ipaths_get() do
+         # remove the current dir, also skip pasrer due to a unidoc bug
```
Don-Ward, line 105: typo: pasrer

uni/ulsp/database.icn
```diff
@@ -96,14 +96,15 @@ class LSPDB(
...
+         if not((dir == ("." | "")) | find("parser", dir))then
```
Don-Ward, line 106: I think this might be a bit too sweeping. It removes any path with "parser" somewhere in it.  I think I'd prefer "uni/parser"

Don-Ward (changes requested):
I think the following is an improvement
```
     every dir := !ipaths_get() do {
         # remove the current dir, also skip uni/parser due to a unidoc problem
        if not((dir == ("." | "")) | find("uni/parser", dir)) then {
           put(db_paths, dir)
           write("   ", dir)
        }
      }
```

uni/ulsp/database.icn
Jafaral: We print it down below, but there was a duplicate "Looking in" that I took out now.

uni/ulsp/database.icn
Jafaral: fixed

uni/ulsp/database.icn
Jafaral: fixed

StephenWampler:
Is this new behavior with UniDoc and parser.icn or was parser.icn never passed to UniDoc before?  I can take a look at it from the UniDoc to see why it's unhappy...

PR #478, opened 2024-09-17
Fix install of ulsp.

Don-Ward:
Adding the ulsp directory to /usr/local/lib/unicon fixes a crash in build_bnf() at start-up.


Jafaral (approved):
We probably shouldn't install ulsp sources since these are not library functions, but the works as a work around for now.

PR #479, opened 2024-09-18
Fix crash in ulsp, additional logging and cleanup

Jafaral:


uni/ulsp/server.icn
```diff
@@ -379,6 +389,7 @@ class Server(
 
    method build_response(id, results)
       local responseBody, responseHeader
+      /results := "{}"
       responseBody := "{\"jsonrpc\":\"2.0\",\"id\":"||id||",\"result\":"||results||"}"
```
Jafaral, line 393: Assuming `result` is always set as a string was the cause of multiple crashes. If `result` is `&null`, just set to an empty object.

PR #481, opened 2024-09-21
ulsp: move the logger to a separate file

Jafaral:


Don-Ward (changes requested):
I'm guessing that this is in preparation for a future move to a separate library class.

Is it too early to set a good example and put some unidoc descriptions in?
At the very least, we should add
```
# To do:
#     Add unidoc descriptions
#     Make the class thread-safe
```
to the top of the `Logger` class.

Names of things:  I think that there is a potential for an unfortunate confusion in names like `debug_class` and `add_debug_class`. Is it a Unicon class we are debugging or adding or a "class" of debug output? It might be better to substitute "category" (or a similar word) for "class".

Jafaral:
@Don-Ward  wish granted

PR #486, opened 2024-10-22
Document the fourth parameter of spawn() in the language reference

Don-Ward:
Also clarify that the defaults are based on the corresponding sizes for the main thread (rather than all being based on heap size).


Jafaral:
Isn't that the forth parameter?

PR #488, opened 2024-10-26
Fix bug in u_read()

Don-Ward:
It was possible to overrun the strings region leading to corruption of memory. The new code checks for overrun and allocates a new buffer if it would occur.

### DIFF
--- a/src/runtime/rposix.r
+++ b/src/runtime/rposix.r
@@ -2445,9 +2445,21 @@ dptr u_read(dptr f, int n, int fstatus, dptr d)
          /* Something is available: allocate another chunk */
          if (i == 0)
             StrLoc(*d) = alcstr(NULL, bufsize);
-         else
+         else {
             /* Extend the string */
+           /* We must guard against running over the end of the current string region.
+            * In that case, allocate a whole new buffer (which will result in a GC)
+            * and copy the existing buffer into it. Don't use alcstr() to do the copy
+            * because that might involve accessing potentially non-existent memory after
+            * the end of the (old) string region.
+            */
+           if (DiffPtrs(strend,strfree) < bufsize) {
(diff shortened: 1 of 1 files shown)


src/runtime/rposix.r
```diff
@@ -2445,9 +2445,22 @@ dptr u_read(dptr f, int n, int fstatus, dptr d)
...
+         else {
             /* Extend the string */
+           /* We must guard against running over the end of the current string region.
+            * In that case, allocate a whole new buffer (which will result in a GC)
+            * and copy the existing buffer into it. Don't use alcstr() to do the copy
+            * because that might involve accessing potentially non-existent memory after
+            * the end of the (old) string region.
+            */
+           int newsize = StrLen(*d) + bufsize;
+           if ( strfree < (StrLoc(*d) + newsize)) {
```
Jafaral, line 2457: I suggest A slight optimization by not doing the work unless we actually need to, since in the vast majority of cases we wont be using `newsize`:
```
           if  (strend-strfree)< bufsize  {
                           int newsize = StrLen(*d) + bufsize;
```

Jafaral (changes requested):

src/runtime/rposix.r
Jafaral: Of course you can just skip `newsize` altogether and just do 
```
char *newb = alcstr(NULL, StrLen(*d) + bufsize );
```

Don-Ward: OK, will do.  I suspect that any self respecting compiler will emit almost the same code for either alternative, but it's probably better not to give the compiler the opportunity to not optimise it. 
In other places, I've noticed we use 

> `if (Diffptrs(strend,strfree) < bufsize)`

so I've gone with that for consistency.

Jafaral (approved):

PR #489, opened 2024-11-08
ulsp: add a json rpc lib, including  a jrpc test 

Jafaral:


uni/ulsp/server.icn
```diff
@@ -23,98 +23,78 @@ class Server(
...
+      params := msg.get_params()
+      id := msg.get_id()
+
+      Logger.info("jrpc:request: ", meth, " id=", id)
+      Logger.debug5_kind("jrpc:request:params ", ximage(params))
+
+      meth ?
+         if ="$/" then {
+            # per LSP spec, this is a protocol implementation dependent request.
+            # we can ignore it. However, we need to still send back a response: TODO
```
Don-Ward, line 42: Should the TODO be removed ?  It appears as if the situation has been covered.

uni/ulsp/jsonrpc-test.icn
```diff
@@ -0,0 +1,107 @@
...
+         }
+      }
+   end
+
+   method handle_request(msg)
+      local id := msg.get_id(),
+            x := integer(msg.get_params()[1]),
+            y := integer(msg.get_params()[2])
+
+      # uncomment to handle potentially trigger some test failures
```
Don-Ward, line 30: There is meaning struggling to escape from this comment, but I can't quite work out what it is :-)

Don-Ward (changes requested):
Minor comments about comments plus a suggestion which can be ignored if you choose.

PR #490, opened 2024-11-13
Add a ThreadPool class to the threads package.

Don-Ward:
The new class is documented in Appendix C of the Unicon book. A test has been added to tests/threads.


Jafaral:
I have't gotten a chance to review this yet. However, it is good to have separate commits for tests and documentation updates. This PR can be split to 3 commits.

Don-Ward:
> I have't gotten a chance to review this yet. However, it is good to have separate commits for tests and documentation updates. This PR can be split to 3 commits.

Any commit that adds something new can be split up into "Add a thing", "Document the thing" and "Test the thing", but it is completely unclear to me why that is better. Splittting it up means that individual commits can be reverted, but why would you revert the addition without also reverting the testing or documentation? In short, I don't understand the benefits of separating documentation and testing from the addition of the thing itself.

Jafaral:
Any commit that can be split to more logical commits, not only tests/docs, should be spilt to more logical commits. It means that the author did put some effort into the design of the feature and organizing the code into small modular additions. It also makes it easier to review the code because the reviewer can follow those smaller commits.  In 3 years we we discover a bug in this thread pool and we trace it back to this commit, it is easier to look at the actual code changes and not have to swift through 100s of other changes that include docs/tests. I can keep going with more reasons why spiting changes into smaller commits is a good idea.

uni/lib/thread.icn
```diff
@@ -504,6 +504,302 @@ procedure ClosePool()
...
+
+   #--------------------------------------------------------------------------------
+   # Fail if there is work in progress, or waiting to be done.
+   method IsIdle()
+      ::lock(Work)
+      if ((Idlers ~= *\Workforce) | (*ToDo > 0)) then { ::unlock(Work); fail }
+
+      ::unlock(Work)
+      return   # success
+   end
```
Jafaral, line 697: Practically the same  as (no need to lock)
```
(Idlers = *\Workforce) & (*ToDo = 0) & return 
```

uni/lib/thread.icn
```diff
@@ -504,6 +504,302 @@ procedure ClosePool()
...
+
+   #--------------------------------------------------------------------------------
+   # Shut up shop. Remaining work placed in the queue before the call of ClosePool
+   # will be executed before closure. Fails if all work has not been done.
+   method ClosePool()
+      local n
+      every n := 1 to *Workforce do Dispatch(stopWork)
```
Jafaral, line 704: every !*Workforce do Dispatch(stopWork)

uni/lib/thread.icn
```diff
@@ -504,6 +504,302 @@ procedure ClosePool()
...
+      every ::wait(!Workforce)
+
+      # At this point, Idlers must be zero.
+      if Idlers > 0 then ::stop("(ClosePool) Idlers queue not empty")
```
Jafaral, line 708: library functions shouldn't exit, generally speaking.

Jafaral (changes requested):
First round of comments. 

I generally don't like the extra braces around expression in contexts when there is no room for ambiguity. I think the code becomes cluttered with unnecessary braces making  it less readable.

Jafaral:
Separating test and documentation doesn't mean you can't do them as you write the actual code.  You do them all at the same time. When you commit,  you just group code together, testing together, documentation together. The code change commit comment describes the code change. Then the test commit does the same, followed by documentation. It is all one PR. If I look at the PR I can see the that you did all three. I can review each stage separately. 

You never have to wonder about what changed when you use git. When I edit a line in my editor, I know who changed the line, and when, and in what PR. with a hover I know what commits exist in the PR and if a test and/or documentation came with that code change on separate commits even before I have to browse the actual changes of the commits. I can jump to the documentation commit right away if it is there.

In all cases, this is just a suggestion given the lessons I learned from my experience working on several projects. Feel free to ignore if you don't see the value as I see it.

PR #498, opened 2025-01-05
runtime: fix null pointer deference on a window close

Jafaral:
We can't queue an event to a window after closing it. Move the queue call before we call wclose().
Simplify the code a bit in the process.

### DIFF
--- a/src/runtime/rxwin.ri
+++ b/src/runtime/rxwin.ri
@@ -831,19 +831,20 @@ int handle_misc(wdp wd, wbp w)
              */
             struct descrip d;
             int ret = 0;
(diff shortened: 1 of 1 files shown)


Don-Ward (approved):
It's not obvious from the diff listing that the change has the same effect as the original (except fixing the queue event on a closed window bug). But after reading the earlier code -- which sets `ws = wb->window;` it becomes clearer.

On a personal note, I think
```
   if (expr)
      return 1;
```
is actually worse than   `if (expr) return 1;` but we could argue all day about coding conventions and still get nowhere (and it's not important enough to ask for changes).

PR #501, opened 2025-02-03
Add support for FreeBSD build

Jafaral:


Don-Ward (changes requested):
It looks like more work is required. In particular, bsd make does not support ifeq.

Jafaral:
> It looks like more work is required. In particular, bsd make does not support ifeq.

fixed!

PR #504, opened 2025-02-16
Added LSIF capability to ULSP

nshapiro4:
Adds lsif.icn into the uni/ulsp directory. This works as a command option in the ulsp command. Can be used to generate LSIF information from a specific directory, or all directories targeted by ulsp.

### DIFF
--- a/src/runtime/rposix.r
+++ b/src/runtime/rposix.r
@@ -1259,7 +1259,7 @@ int sock_listen(char *addr, int is_udp_or_listener, int af_fam)
 
 
    if ((s = sock_get(addr)) < 0) {
-     char *p;
+     char *p, fname[BUFSIZ];
 
      /*
       * If the first argument is just a name, it's a unix domain socket.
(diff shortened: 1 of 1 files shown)


Jafaral:
You have a merge commit, you can please rebase on top of master?

jschnet:
The `output.lsif` file probably shouldn't be part of the source, unless told otherwise. This can be removed by adding a line to the [.gitignore](https://github.com/uniconproject/unicon/blob/master/.gitignore#L89) file instead, right after that ulsp line. 

Also would be good to remove this file while cleaning using the [Makefile](https://github.com/uniconproject/unicon/blob/master/uni/ulsp/Makefile#L35), i.e. `make clean`. 

If we anticipate there being more than just this one output file in the future, having a directory called `outputs/` in here would make things easier since there would be less to add for instances like this since we can just add that directory to the .gitignore and Makefile rather than each individual file.

PR #509, opened 2025-02-26
runtime: don't change string in place

Jafaral:


Don-Ward (approved):
Looks good to me.  I wonder if this happens anywhere else.

PR #511, opened 2025-03-12
Enable a few tests

Jafaral:


Jafaral:
I had the same thought. This is a new testing paradigm that I would like to make use of. I'd like to split this into a new directory and add more tests using the framework. In the short term, I at least wanted to make sure there is some test coverage for it. We already have support for ignoring differences in output, which I used in this case to allow the tests to pass if you noticed. By having it run, we can catch issues with visual inspection of the test output at least, even on the CI side.

Jafaral:
@Don-Ward  in summary, the test passed the CI on all platforms as part of this PR. So this is not an issue.

PR #512, opened 2025-03-13
Add symbol and folding request handling

nshapiro4:
Added support for textDocument/documentSymbol and textDocument/foldingRange requests in ULSP. Modified filehandler.icn to store ranges of symbols for use in other handlers.

### DIFF
--- a/uni/ulsp/file_handler.icn
+++ b/uni/ulsp/file_handler.icn
@@ -1,5 +1,5 @@
 # -------------------------------------------------------------------------
-# ---------------------- FileHandler Class ----------------------------------
+# ---------------------- FileHandler Class --------------------------------
 # -------------------------------------------------------------------------
 #
 # Description:
@@ -50,7 +50,8 @@ class FileHandler(
    package_classes,
    m_uni_file,
    temp_idoc,
-   context
+   context,
+   internal_positions
    )
 
    method updateFileHandler()
@@ -67,6 +68,7 @@ class FileHandler(
       setPackageClasses()
       setObjects()
       setVariables()
+      setInternalPositions()
    end
 
    method setFile()
@@ -247,7 +249,7 @@ class FileHandler(
    end
 
 # -------------------------------------------------------------------------------------------------
-# ---------------- Package Procedures (get, set, update) -----------------------------------------
+# ---------------- Package Procedures (get, set, update) ------------------------------------------
 # -------------------------------------------------------------------------------------------------
 
    method setPackageProcedures()
@@ -377,7 +379,7 @@ class FileHandler(
    end
 
 # -------------------------------------------------------------------------------------------------
-# ---------------- Package Classes (get, set, update) --------------------------------------------
+# ---------------- Package Classes (get, set, update) ---------------------------------------------
 # -------------------------------------------------------------------------------------------------
 
    method setPackageClasses()
@@ -427,7 +429,7 @@ class FileHandler(
    end
 
 # -------------------------------------------------------------------------------------------------
-# ---------------- Variables (get, set, update) -----------------------------------------------------
+# ---------------- Variables (get, set, update) ---------------------------------------------------
 # -------------------------------------------------------------------------------------------------
 
    method setVariables()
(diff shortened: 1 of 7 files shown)


uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+            if ReMatch("procedure\\s+" || procedureName || "\\s*\\(", _line) then {
+               startPos := find(procedureName, _line)
+               endPos := startPos + *procedureName - 1
+               push(open_blocks, ["procedures", procedureName])
+               internal_positions["procedures"][procedureName] := table(
+                  "type", "procedure",
+                  "range", table(
+                     "start", [lineNum-1, 0],
+                     "end", [lineNum-1, 0]
+                  ),
```
Jafaral, line 519: table syntax slightly more readable ` := [ K : V; ... ]`

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+                  "selectionRange", table(
```
Jafaral, line 520: ditto

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+               internal_positions["classes"][className] := table(
+                  "type", "class",
+                  "range", table(
+                     "start", [lineNum-1, 0],
+                     "end", [lineNum-1, 0]
+                  ),
+                  "selectionRange", table(
+                     "start", [lineNum-1, startPos],
+                     "end", [lineNum-1, endPos]
+                  )
```
Jafaral, line 545: more

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+                     "type", "method",
+                     "range", table(
+                        "start", [lineNum-1, 0],
+                        "end", [lineNum-1, 0]
+                     ),
+                     "selectionRange", table(
+                        "start", [lineNum-1, startPos],
+                        "end", [lineNum-1, endPos]
+                     )
+                  )
```
Jafaral, line 565: more

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+               "type", "method",
+               "range", table(
+                  "start", [lineNum-1, 0],
+                  "end", [lineNum-1, 0]
+               ),
+               "selectionRange", table(
+                  "start", [lineNum-1, startPos],
+                  "end", [lineNum-1, endPos]
+               )
+            )
```
Jafaral, line 586: this too

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+         }
+
+         #Check for keyword "end"
+         if ReMatch("^\\s*end\\s*$", _line) then {
+            if *open_blocks > 0 then {
+               ending := pop(open_blocks)
+               if ending[1] == "initially" then {
+                  internal_positions[ending[1]][ending[2]]["range"]["end"] := [lineNum-2, *temp_idoc.getUniFileLine(uri, lineNum-1)]
+                  ending := pop(open_blocks)
+                  internal_positions[ending[1]][ending[2]]["range"]["end"] := [lineNum-1, *_line]
```
Jafaral, line 596: line 596 overwrites 594? was this left here accidentally?

uni/ulsp/folding.icn
```diff
@@ -0,0 +1,120 @@
...
+               if startLine < endLine then {
+                  put(results_table, table(
+                     "startLine", startLine,
+                     "endLine", endLine
+                  ))
+               }
+            }
+         }
+      }
+      every result := !findFoldingRanges() do put(results_table, result)
```
Jafaral, line 41: you can eliminate the loop variable: 
```
every put(results_table,  !findFoldingRanges())
```

Jafaral: Looking again at `findFoldingRanges()`, looks like it returns a list. In this case you just want to concatenate the two lists:
```
results_table  := result_table ||| findFoldingRanges()
```
or just
```
results_table  |||:= findFoldingRanges()
```

uni/ulsp/server.icn
```diff
@@ -251,9 +271,51 @@ class Server(
...
+#                textDocument/documentSymbol           #
+########################################################
 
-   method shutdown()
+   method symbolHelp(msg)
+      local filehandler_object, result
+      filehandler_object := openFiles[\msg.get_param_uri()]
+      symbolHandler.setVariables(filehandler_object)
+      if renewSymbols = 1 then {
+         Logger.info("Getting new symbols for " || \msg.get_param_uri())
```
Jafaral, line 283: If you want to allow this call to always succeed, then 
```
 Logger.info("Getting new symbols for " || (\msg.get_param_uri() | "")
```
However, given that you don't store the result if we don't have a uri, might as well wrap this black with a check if we have uri or not, so it becomes:
```
if \msg.get_param_uri() then }
   Logger.info("Getting new symbols for " || msg.get_param_uri())
   result := symbolHandler.run()
   cacheSymbols[msg.get_param_uri()] := result
}
```
Also, these logs seem like debug msgs rather than info.

uni/ulsp/symbols.icn
```diff
@@ -0,0 +1,202 @@
...
+            )
+            entity_table[procedure_name] := proc_symbol
+            put(results_list, proc_symbol)
+         }
+      }
+
+      # Get classes and their methods from filehandler
+      every class_name := key(filehandler.getInternalClasses()) do {
+         if member(pos_table["classes"], class_name) then {
+            class_table := pos_table["classes"][class_name]
```
Jafaral, line 52: instead of doing two table lookups, you can do:
```
if  /(class_table := pos_table["classes"][class_name]) then {
   ...
```

uni/ulsp/symbols.icn
```diff
@@ -0,0 +1,202 @@
...
+               class_table["selectionRange"]["end"][1],
+               class_table["selectionRange"]["end"][2],
+               []
+            )
+         }
+
+         # Get methods inside classes from filehandler
+         every method_name := key(filehandler.getInternalClasses()[class_name]["methods"]) do {
+            if member(pos_table["methods"], method_name) then {
+               method_table := pos_table["methods"][method_name]
```
Jafaral, line 72: ditto: one table lookup

uni/ulsp/symbols.icn
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      # Get variables from filehandler
+      every variable_name := key(filehandler.getVariables()) do {
+         if member(pos_table["variables"], variable_name) then {
+            variable_table := pos_table["variables"][variable_name]
+            sl := variable_table["selectionRange"]["start"][1]
+            sc := variable_table["selectionRange"]["start"][2]
+            el := variable_table["selectionRange"]["end"][1]
+            ec := variable_table["selectionRange"]["end"][2]
```
Jafaral, line 123: In this case and all similar code blocks above and below, you gotta decide if it is cheaper/faster to store the first table  lookup in a local variable instead of doing it multiple times.  
```
selection_range := variable_table["selectionRange"] 
sl := selection_range ["start"][1]
...
```
And, at least in this code block , I don't see any use for `variable_table` outside these lookups, so you may want to just skip `variable_table` altogether and go directly to the item you care about, so the above code becomes : 
```
selection_range := pos_table["variables"][variable_name]["selectionRange"] 
sl := selection_range ["start"][1]
...
```

uni/ulsp/symbols.icn
```diff
@@ -0,0 +1,202 @@
...
+   #  Params: v_rsl - The variable's starting line of its range.
+   #          v_rsc - The variable's starting character of its range.
+   #          v_rel - The variable's ending line of its range.
+   #          v_rec - The variable's ending character of its range.
+   #          entity_table - A table containing all entities in the document, i.e. classes, methods, procedures.
+   #  <i>This is intended for internal use only!</i>
+   method find_parent_entity(v_rsl, v_rsc, v_rel, v_rec, entity_table)
+      local parent, p_rsl, p_rsc, p_rel, p_rec
+      every parent := !entity_table do {
+         p_rsl := parent["range"]["start"]["line"]
```
Jafaral, line 147: I think this works too
```      
every range := (parent := !entity_table)["range"] do {
         p_rsl := range["start"]["line"]
```

uni/ulsp/symbols.icn
```diff
@@ -0,0 +1,202 @@
...
+               "line", ssl,
+               "character", ssc
+            ),
+            "end", table(
+               "line", sel,
+               "character", sec
+            )
+         ),
+         "children", chi
+      )
```
Jafaral, line 198: table syntax `[K : V; ...]`

uni/ulsp/file_handler.icn
nshapiro4: Line 596 is after popping open_blocks, giving a new value to ending. I did it this way to pop the "initially" section off the stack, put it in the table, and then pop the next item which would be the class. So line 594 would be internal_positions["initially"][className] while line 596 would be internal_positions["classes"][className]

uni/ulsp/file_handler.icn
```diff
@@ -486,6 +488,140 @@ class FileHandler(
...
+
+   method setInternalPositions()
+      local lineNum, _line, procedureName, startPos, endPos, className, methodName, variableName,
+            open_blocks, ending, currentClass
+
+      internal_positions := [
+         "classes" : [];
+         "procedures" : [];
+         "methods" : [];
+         "variables" : []
```
Jafaral, line 503: These will end up being lists now, not tables since `[ ]` is a list initializer syntax.  To force it to be a table you can do  `[:]` which is short of `[&null: &null]` though it is common to just use `table()` as you did before.

uni/ulsp/server.icn
```diff
@@ -251,9 +271,54 @@ class Server(
...
    end
 
+########################################################
+#                textDocument/documentSymbol           #
+########################################################
+
+   method symbolHelp(msg)
+      local filehandler_object, uri, result
+      
+      if uri := msg.get_param_uri() then {
```
Jafaral, line 281: This will always succeed. So I think you still need a no null check ` \(uri := msg.get_param_uri()) `

Jafaral (approved):

PR #513, opened 2025-03-20
Add log level change notification to ULSP

nshapiro4:


uni/ulsp/logger.icn
```diff
@@ -321,6 +321,10 @@ class Logger(
    method debug_level_kind(lvl, kind, msgs[])
       self._debug(lvl+8, kind, msgs)
    end
+   
+   method change_log_level(newLevel)
+      current_log_level := newLevel
+   end
```
Jafaral, line 327: This is a typical `set` function, let is call it `set_log_level()`

Jafaral (approved):

PR #515, opened 2025-04-10
Fix ivib to cope with changes in commit 7d4093ae.

Don-Ward:
7d4093ae changed the compute_absolutes() method in the Dialog class. A side effect of this change is that the coordinates of ivib's main dialog were no longer set to (0,0) before the dialog is displayed by custom_show(), which led to extra space in the dialog window (and an ivib crash under some circumstances).

The smallest, and probably safest, fix is to set the coordinates explicitly in the custom_show() method.

### DIFF
--- a/uni/ide/ivib/main.icn
+++ b/uni/ide/ivib/main.icn
@@ -286,6 +286,7 @@ class Main : CommonDialog(
    method custom_show(a)
       local s, f, c
       self.args := a
+      self.set_abs_coords(0,0)
       show()
       dispatcher.add(self)
       if *self.args > 0  then {


StephenWampler:
There are a lot of subclasses to Dialog that don't override compute_absolutes - have they all been checked to make sure they 'pop-up' correctly with the change to compute_absolutes?  If only Main shows the problem then I have no issue with the proposed fix.

Don-Ward:
> There are a lot of subclasses to Dialog that don't override compute_absolutes - have they all been checked to make sure they 'pop-up' correctly with the change to compute_absolutes? If only Main shows the problem then I have no issue with the proposed fix.

Applications built with ivib work normally (both before and after 7d4093ae and also after this fix). The problem is with ivib itself: without the fix, the main menu bar and the canvas are offset vertically. It is possible to increase that vertical offset by selecting `Canvas->Dialog Prefs` and then cancelling. After a few cycles, the increased vertical offset places the canvas outside the dialog window and ivib crashes.

Even with the vertical offset, the components are placed in the correct position in the canvas (and correctly placed in the target application).

The possible issue with 7d4093ae (and the change of behaviour in `compute_absolutes`) is not with applications built with ivib, but with other applications (such as ivib itself) that use the gui classes directly and depend on the old behaviour. My assumption is that in the nearly five years since 7d4093ae they would already have been adjusted to cope with the new behaviour if it made any difference. Either that or there might have been complaints that we have changed the gui classes and broke someone's program. So changing `compute_absolutes` back risks breaking more applications. Hence the decision to change ivib itself. Before 7d4093ae, `compute_absolutes()` didn't change since its introduction in 2003. Ivib was introduced a year later. It's impossible to know at this distance whether the side effect on the absolute position was intentional or accidental; i.e. whether ivib was written to rely on that behavour or whether it just happened to work and the reliance on the side effect went unnoticed.

We have been aware of the change in ivib's behaviour for a couple of years or more. It's only now that someone has been motivated enough to find the root cause.

Jafaral (approved):

PR #525, opened 2025-08-14
Fix typo in m_write() method.

StephenWampler:
There's a typo in m_write() that causes it to produce bad results.


Jafaral (changes requested):
@StephenWampler , can you please rebase this branch on top of latest master?

Jafaral (approved):

PR #526, opened 2025-09-01
Fix a ULSP crash and improve code readability a bit 

Jafaral:


uni/ulsp/server.icn
```diff
@@ -175,7 +175,9 @@ class Server(
          }
       }
       # Get rid of root / and change unicode encoded "%3A" back to ":" on Windows machines
-      fname := replace(fname[2:0], "%3A", ":")
```
Jafaral, line 178: On non Windows systems, the latest PR added this code that is  stripping the leading `/` character causing ULSP to fail to find any source file.

PR #527, opened 2025-09-01
Autoconf update and clean up of obsolete macros 

Jafaral:


Jafaral:
@Don-Ward I came across these while trying to make #521 work with autoconf help

PR #531, opened 2025-09-12
Add a new keyword &proc to access the Unicon function proc()

brucerennie:
I have cloned a copy of the current uniconproject/unicon repository and updated keyword.r, kdefs.h, keyword.h and unigram.y to include a new keyword &proc.

Jafar, in the past you had a number of additional keywords that you wanted to include, do you have the details for those new keywords?

Notes and solution form previous pull request:

*******************************************************************************************************

Something is quite odd here. I have just added the same code that was submitted a couple of years ago. However, it has adversely affected the keyword &time in such a way that &time now returns "" as its result and not an integer.

In the original submission, the code was added to the end of the file keyword.r. In the testing on my machine, I have shifted the code to other places in the file keyword.r and still have the same error.

As far as I can see, there should be no interference between &time and &proc.

*****************************
SOLUTION
*****************************

Okay, Now that I have done some testing, it appears the problem is related to mkkwd.icn.

If I remove the contents of both kdefs.h and keyword.h and then run mkkwd, it doesn't then matter where you put the new keyword code in the keyword.r file.

The values associated with the K_ definitions will be recreated and the new keywords will be placed in the proper alphabetic sequence with continuously increasing numeric values.

Using the current official unicon files kdefs.h and keyword.h files, new keywords added will get a defined value of K_* larger than the existing ones, but this new value is not in sequence in the keyword.h file.

In my original submission from a few years ago, I added the &proc code to the end of the keyword.r and manually updated both kdefs.h and keyword.h to put the relevant entry at the end of each file


Jafaral:
The PR replaced #530

Jafaral:
Thanks @brucerennie. If you add these keywords, you can simply have them fail. I can add the implementation later.

brucerennie:
Jafar,

The &feature item "Libraries at" is in the code but is never returned. It appears to be related to the following code:

      if ((int)strlen(uniroot) > 18) {
         char *s;
         s = alcstr(NULL, strlen(uniroot+18) + strlen("Libraries at ") + 1);
         strcpy(s, "Libraries at ");
         strcat(s, uniroot+18);
         suspend C_string s;
         }
 
There is another problem in that the code for "Binaries at" always returns the bin directory where the original system is compiled, in this particular case /home/bruce/src/git/github/unicon-official/bin

This is the value returned even when the binaries are placed elsewhere /home/bruce/.unicon/bin and even after the patchstr command is run. I am not sufficiently au fait with the code at this point to track down why.

I have separated out the code relevant to &config. However, when I remove that specific code from &features either by actually removing the lines or commenting them out, when compiling the xml libraries and error arises where it can no longer find the package lang which is found in the uniclass db in the lib directory. 

At this point, this is stopping me from submitting the relevant patches. &child and &parent work by failing.

Also, as reported above, the files keyword.h and kdefs.h both have to be cleared for the additional keywords can compile correctly. Using the mkkwd.icn file to regenerate these files.

brucerennie:
Jafar,

The &feature item "Libraries at" is not actually initialised prior to &features keyword being called. This is why it is not being returned. The variable uniroot only has the "PatchUnirotHere ->" value in it. Refpath is initialised to the correct location required.

In addition if you remove some of this code from &features code, it messes with the IPATH and/or the LPATH variables and the compilation process fails. In effect even changing the return statements to fails causes the compilation of the unicon libraries to fail. It initially showed up when compiling the xml library because it couldn't find the uniclass db. Now some changes cause the inability to find include files.

Restoring the code for &features eliminates the problems. The additional keywords are not causing problems here.

At this point the particular things that should be removed from &features, which are now in &config, are having some strange interactions with the rest of the code for icont/iconx.

I am still investigating and will let you know what I find further.

PR #596, opened 2026-04-15
Add an option to disable native coswitch,  pthread coswitch fixes

Jafaral:


cjeffery:
While looking at this, I saw mention of ulimit being needed under some circumstance. I think we have some C stack size-setting code (in our C code, presumably init.r or some such) on platforms where that is available, can we review that and see if the defaults there should be increased? Similarly vis-a-vis the recent apparent need for larger MSTKSIZE and/or STKSIZE, should those defaults be increased?

Jafaral:
> While looking at this, I saw mention of ulimit being needed under some circumstance. I think we have some C stack size-setting code (in our C code, presumably init.r or some such) on platforms where that is available, can we review that and see if the defaults there should be increased? Similarly vis-a-vis the recent apparent need for larger MSTKSIZE and/or STKSIZE, should those defaults be increased?

I'm only seeding an intermittent failures on FreeBSD, and it rarely happens. The fix really target only that run. Though I agree that we can revisit C stack  resizing. Will see what I can do for those later.

PR #604, opened 2026-04-30
Teach RTT to parse some new runtime C constructs

Jafaral:


Jafaral:
@cjeffery , wish granted - from #603

cjeffery:
This is pretty darn awesome.  The main weirdness for me is the %token declarations for integer literals e.g. 257 interspersed with the actual token names.  I wish there was a comment to explain why that is required.

The gigantic changes in generated code -- did we switch from yacc to bison at this point?  I guess there are downsides to including those in the repository, but they are there for bootstrapping purposes and I guess we live with it.

Don-Ward (approved):
Looks good to me. My only comment is about licenses. This is the first (I believe) example of code that requires GPLv3 (or later). Everything else is GPLv2 (or later). If we're happy with that then it's good to go. If not, we might be able to find an earlier version of Bison  that is licensed under GPLv2 or GPLv2.1 and use that instead.

Jafaral:
> Looks good to me. My only comment is about licenses. This is the first (I believe) example of code that requires GPLv3 (or later). Everything else is GPLv2 (or later). If we're happy with that then it's good to go. If not, we might be able to find an earlier version of Bison that is licensed under GPLv2 or GPLv2.1 and use that instead.

The Bison tool itself is GPLv3, but we don't ship Bison — we ship its output. As long as the generated files carry the exception notice (which Bison 3.8.2 does), we're fine with Unicon's GPLv2/2.1 license.

src/rtt/ltoken.h
```diff
@@ -1,117 +1,229 @@
...
+   under terms of your choice, so long as that work isn't itself a
+   parser generator using the skeleton or a modified version thereof
+   as a parser skeleton.  Alternatively, if you modify or redistribute
+   the parser skeleton itself, you may (at your option) remove this
+   special exception, which will cause the skeleton and the resulting
+   Bison output files to be licensed under the GNU General Public
+   License without this special exception.
+
+   This special exception was added by the Free Software Foundation in
+   version 2.2 of Bison.  */
```
Jafaral, line 32: @Don-Ward FYI

Jafaral:
@cjeffery  I added a comment [here](https://github.com/uniconproject/unicon/pull/604/changes/3643f2f51a85bc9291dd035bb30d8edfa29486e3)

Jafaral:
> This is pretty darn awesome. The main weirdness for me is the %token declarations for integer literals e.g. 257 interspersed with the actual token names. I wish there was a comment to explain why that is required.

Not required per se, Bison allows `%token ... Name N Name2 N2 ...` as you know so each symbolic token is paired with its numeric code on the same line (e.g. Identifier 257 StrLit 258 assigns 257 to Identifier and 258 to StrLit).

 
> The gigantic changes in generated code -- did we switch from yacc to bison at this point? I guess there are downsides to including those in the repository, but they are there for bootstrapping purposes and I guess we live with it.

They are gonna sit there mostly unnoticed! and yes, no need to install bison unless the grammar changes obviously.

PR #606, opened 2026-05-01
Add C11 thread_local, _Atomic to rtt and add compiler TLS for curtstate

Jafaral:
The reason for doing this is to get access to compiler-supported concurrency features. Here is some numbers from a quick local benchmark:
```
mode                           iter (M)         time_s      Miter/s                                                                                                
------------------ -------------------- -------------- ------------                                                                                                
pthread_key                         100       0.098692      1013.26                                                                                                
compiler_tls                        100       0.071966      1389.55                                                                                                
atomic                              100       0.346529       288.58                                                                                                
mutex                               100       0.776938       128.71                                                                                                
pthread_key/compiler_tls: 1.371                                                                                                                                    
mutex/atomic: 2.242    
```
Compiler TLS is about 30% faster compared with pthread TLS.  Atomic operation is more than 2 times faster than a pthread mutex. Those features weren't widely available when we introduced concurrency to Unicon. Now they are commonplace.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,6 +95,9 @@ jobs:
     - name: Test
       run: make Test
 
+    - name: TLS Benchmark
+      run: make -C tests/rtt tls_bench
+
     - name: TestReport
       if: ${{ steps.build.outcome == 'success' }}
       run: |
@@ -152,6 +155,9 @@ jobs:
 
     - name: Test
       run: make Test
+
+    - name: TLS Benchmark
+      run: make -C tests/rtt tls_bench
 
     - name: TestReport
       if: ${{ steps.build.outcome == 'success' }}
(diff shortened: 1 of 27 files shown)


Don-Ward (changes requested):
If we are changing the minimum required C standard from C99 to C11, we should change the top level `README.md`. 
Is it worth making some allowance for C23 deprecating `_Thread_local `in favour of `thread_local `(e.g. by making a `THREAD_LOCAL` macro)?  OTOH, we probably won't change to C23 any time soon, so there is no pressing need.
Somewhat to my surprise, these changes do not require any changes to the Implementation Book (that I can see). Perhaps they ought to: should we include a description of how the various data structures in the runtime are used to implement concurrency along with a diagram giving "the big picture" of how it all ties together?

Jafaral:
> If we are changing the minimum required C standard from C99 to C11, we should change the top level `README.md`.

This PR doesn't change the minimum required C standard. C11 new features are all under auto-detect configuration. If they are not available, they don't get used.

 Is it worth making some allowance for C23 deprecating `_Thread_local `in favour of `thread_local `(e.g. by making a `THREAD_LOCAL` macro)? OTOH, we probably won't change to C23 any time soon, so there is no pressing need. Somewhat to my surprise, these changes do not require any changes to the Implementation Book (that I can see). Perhaps they ought to: should we include a description of how the various data structures in the runtime are used to implement concurrency along with a diagram giving "the big picture" of how it all ties together?

It doesn't hurt to detect thread_local. I at least can add it to rtt so it doesn't need a passthru if we use it later.

PR #608, opened 2026-05-07
    test: Remove configuration specific values from mutex test

Don-Ward:
Some configurations have a different number of mutexes, which leads to a test failure (because the test output is not the same as the expected result). This commit changes the output so that exact values are no longer part of the output. The actual tests are unchanged.


Jafaral:
@greptile review

Jafaral (approved):

PR #612, opened 2026-05-19
Fix grammar comments for operator symbols RCV & RCVBK

brucerennie:
The operators RCV and RCVBK previously were @< and @<< in various comments and strings.

Theses have been updated to the correct forms 
RCV <@
RCVBK <<@

in both comments and string values.

In the previous PR related to this (now closed), I made a mistake in applying an update. This PR replicates the previous PR but based on the current uniconproject/master branch.


Jafaral:
@greptile review

Issue #622, opened 2026-07-15
Make fmonitr.r compile on OpenBSD


marcespie:
the profil(3) API is substantially different.

This relies on "Basic support for OpenBSD" to bring in the OpenBSD define.

[diff5.txt](https://github.com/user-attachments/files/30044174/diff5.txt)

Jafaral:
Fixed in Fixed in #633

PR #633, opened 2026-07-17
Use OpenBSD profil(2) signature in fmonitr.r

Jafaral:
From marcespie's patch on uniconproject/unicon#622.


marcespie:
100% agree. Relies on OpenBSD being defined, but that will happen eventually (soon hopefully)

PR #642, opened 2026-07-28
Fix out-of-tree dist and rpm to use the source tree

Jafaral:


Jafaral:
@greptile, `--exclude-vcs` and some other options predated this PR. We are assuming gnu environment for deb/rpm which is what this is used for. I'm restoring the simpler form with `--transform` and going with that for now. If we need to make this to work with non GNU tool later we can fix it.

PR #643, opened 2026-07-29
Modernize Unicon GUI with a flat theme

Jafaral:
* Add a pluggable flat look-and-feel as the GUI package default (classic Motif bevels still available)
* Apply flat chrome across widgets and the Unicon IDE / Ivib (dialogs, menus, tabs, scrollbars, checkboxes)


brucerennie:
There is a circular reference between guiprocs.icn and theme.icn when using 

unidep *.icn -f deps.out -nb

Jafaral:
@brucerennie I  fixed that

brucerennie:
Compilation of ui is now successful. However, a number of display issues are evident. I have attached four png files showing what the display looks like. I will also send these through unicon-group message.

Jafaral:
@brucerennie , can you please test and confirm if the issue is resolved? I dropped font changes from this PR.

brucerennie:
Good morning (or at after midnight) Jafar,

The particular problem with the ui interface appears to be corrected. I did notice though that when opening multiple files using ui, that after 9 files opened, the 10th goes from 62 Mbytes to 1.7 Gbytes, at which time I aborted ui so as to not stall my machine.

I will create a bug report for this as I investigate.

Otherwise, all appears to be okay.

PR #649, opened 2026-08-10
Add initial Unicode Support

Jafaral:


Jafaral:
@brucerennie since you did some unicode/utf8 work on the language level, I'd appreciate if you have some feedback on this PR. a preview of the technical report is available here:

https://jafaral.github.io/unicon/doc/utr/html/utr25.html

brucerennie:
Jafar,

I am working through the above PR and in the file fmisc.r, no merge is possible as the line 3125 has a reference to bloblen but is not found in the latest fmisc.r

Similar problem in fstruct.r. The change is required in line 693 but is found in latest sources at line 908.

Fsys.r also has merge problems - function close and can't find File.fname

Rmisc.r also has merge problems 

Rmswin.ri has line 1261 which would appear to need a SetStrLen used

rposix.r has a number of places where SetStrLen might be needed, lines 1132, 1193, 1240, 1270, 4262,

rwindows.r line 587 appears to need SetStrLen

Will continue working through all the changes

Jafaral:
@brucerennie, thank you for the feedback. If you tried to merge this into the master branch to try it, then that might explain the issue. I merged other PRs in recent days which created conflict with this PR. I will resolve those and update the PR.

brucerennie:
Jafar,

Further to the above message, the file oref.r is also merge resistant. 

A suggestion though. As I was looking through the changes for StrLen to SetStrLen, I noticed that in many cases there was associated StrLoc. Sometimes the order was SetStrLen followed by StrLoc and in other places Strloc followed by SetStrLen.

To unify the code (so to speak), I believe that it could be beneficial to have both calls being replaced by a single macro that performs the current two macros. It is only a suggestion though.

Jafaral:
> Jafar,
> 
> Further to the above message, the file oref.r is also merge resistant.

I fixed the conflicts and pushed the update as you were writing the comment. 

> A suggestion though. As I was looking through the changes for StrLen to SetStrLen, I noticed that in many cases there was associated StrLoc. Sometimes the order was SetStrLen followed by StrLoc and in other places Strloc followed by SetStrLen.
> 
> To unify the code (so to speak), I believe that it could be beneficial to have both calls being replaced by a single macro that performs the current two macros. It is only a suggestion though.

We do have that already actually, I will  make a sweep commit to use that where possible.

Jafaral:
@brucerennie I updated a lot of locations to use `MakeStr` which does both `StrLen` and `SetStrLen`

brucerennie:
@Jafaral I have merged the unicode PR into a current copy of unicon. Looking at the tests, I don't see anything untowards. I did a test of the current ui program and it is not unicode aware in terms of the display. I will be doing an additional test tomorrow regarding multibyte codepoints that should be recognised as illegal.

Jafaral:
Thanks @brucerennie. Notice that I had to disable/skip the two utf8 tests you added a while back because they both fail since now built-in functions and string subscription is now utf8 aware. Those tests assumed byte/ascii semantic which is no longer true.

brucerennie:
@Jafaral Not a problem. I am using the uft8_ovld.icn test as a basis to test all the relevant operators and functions. Once done, i will let you know the results as an alternative to your specific unicode tests.

brucerennie:
@Jafaral The function ord(0 doesn't work on unicode codepoints if larger than 8 bits. As well there are a number of methods in the UTF8 class that would be useful for unicode, One tests for invalid sequences. I will continue checking.

Jafaral:
I kept `ord()` and `char()` byte-oriented for backward compatibility. They are used in in bit packing functions. They return different characters for upper ascii characters if the become unicode-aware.  All other functions were backward compatible. So My PR introduces  two new alternative functions that are unicode aware see [here](https://jafaral.github.io/unicon/doc/utr/html/utr25.html#sec-uchar)

This is something we can change, but at least that is what I went with for now.

brucerennie:
@Jafaral I have read the utr25 and I am happy with the design choices made. I can live quite happily with what you have implemented. Now I just have to work out what i need to do with the IDE ui to display the unicode codepoints on screen. I will get to that eventually, if it isn't solved by others better suited.

cjeffery:
General comments on PR 649

Overall, the implementation seems clever.
A nice job was done figuring out some of icont's internals.

The claim of this highly desirable feature (Unicode support) having
minimal added performance cost are unsubstantiated. Strings are
central to the runtime system and a full correct implementation
involves extensive changes. It is difficult to ascertain whether
those changes are complete, and it is difficult to believe the claim of
low cost.  Minimally, the Unicon benchmarks should be run with and
without Unicode support and compared. Other evidence giving the user
an idea of the cost(s) of the feature (primarily time, it would seem)
would be welcome.

It is not acceptable to merge this feature without proper #ifdef's
around everything.  If your ifdef is not defined, the original code
should be present and provided unmolested.  If that is the case, all
your new stuff should just not be compiled in, and not be imposing
the many additional runtime checks currently in the PR, Unicode or no.

In utr25.rst, the generous co-author credits should be checked for
co-author permission.  For example, I will claim that I was not
involved and am not a co-author, unless you can convince me
otherwise. Maybe there was an e-mail or two where the prospective
feature was discussed very vaguely, but that does not mean I was
involved in the design that was chosen.

Although not as big an issue as the proper #ifdef's being absent,
turning on Unicode by default raises the bar that this must pass before
the PR should be merged. It is normal for a feature as substantive as
this one to go through a public beta period. Calling it "optional"
when it is on by default is disingenuous. Instructions for how to turn
it off in the configure step would be helpful to non-programmer
users. But perhaps it would be best to start with a period where it is
optional and the user has to explicitly ask for it.

The discussion of whether to cache a "codepoint count" or a "scan
position" is interesting, but confusing.  A data descriptor encoding a
property of the data (codepoint count) seems non-dangerous, but one
encoding a property of a particular computation ("scan position")
sounds like trouble. That is not just data any more.

There is an interesting question about when to cache codepoint counts,
as described in the discussion of concatenation. One can describe the
current implementation as lazy (in the respectable CS programming languages
sense) but possibly deserving some tweaking; an alternative approach might
be not quite so conservative about codepoint counting. If the information is
computable at compile time, it is a shame to leave it to runtime.  And at runtime,
a lot of large strings get constructed incrementally by adding one character
or one line at a time to an existing (increasingly large) string. If a
concatenation of a large string with counted codepoints concatenated onto
a small string without counted codepoints results in discarding a large
computed result, it might seem cleverer to incrementally update the existing
codepoint count at concatenation time. The threshold where it becomes
logical to discard known codepoint count information rather than computing
an incremental update is a fine question.

The discussion of read() and reads() and similar input initially read as
being "raw only", a puzzling choice. The discussion of "i" mode later
clarified things, and maybe should come earlier. But "i" is a curious mode
letter to use. If it stands for something, what is it? If it doesn't stand
for anything mnemonic, we should consider alternatives.  It is too bad
mode letters seem to be case-insensitive, or we could leave "u" as
untranslated but take "U" for "Unicode". Oh well.

Discussion at second paragraph of 7.0 of the report, right before 7.1, is unclear.
Notion that the way to drop a tag is to write the string and read it back in
again seems silly. Discussion in 7.2 is unclear.

Internally, strings get used a lot for more than just * [] and move/tab/pos.
Is the behavior checked everywhere that the RTL code says is:string
and so forth?  Apparently in many locations, the unicode string is still a
valid icon string and so you can get away with not customizing code, but
there are liable to be surprises.

Similarly, do the various StrLoc(x)++ operations in the runtime need to be
modified to advance one whole codepoint if UniQual, with corresponding
StrLen() modifications, or can we prove they are never UniQual?
See fwindow.r

The work is incomplete vis a vis csets and the string analysis functions
that use them.

You've changed StrLen(d) to use a bitwise AND, and this change appears to
apply whether Unicode is compiled in or not. This is not OK.

For StrLen(d) and SetStrLen(), we need to understand if bitwise-and'ing on
64-bit operations is a faster option than doing direct/native 32-bit operations.
I suppose the answer might vary by CPU type. But for instance in cnv.r you
turn a ++StrLen(d) into a SetStrLen() and I am wondering if a casted 32-bit
++ would still be faster.

One exception to this is: it would be possible to define a
SetStrLen() and introduce its use everywhere needed, separate and prior
to the main PR. The MakeStr() macro could similarly be introduced as part
of the separate/prior PR. But it will be challenging to enforce the new
required discipline in descriptor handling on other developers.

"Unicon Phase 0" language in comments in various files is confusing.
Do you mean Unicode Phase 0?  In fstranl.r it is even worse, just saying
"Unicon:" in places, not even "Unicon Phase 0".

Comments are needed where you are tending new variables in fsys.r --
why is the new tended variable needed in those places? Is it that RTT's
return string(x,y) doesn't know to do the correct thing yet? If so why not?
And why is the new variable tended in these various spots? Are you introducing
additional allocations somehow?

brucerennie:
@Jafaral @cjeffery In your comment above, in regard to unicode csets, as I discovered when writing the UTF8 classes, one has to be careful here as the sizes applicable to the selected unicode cset means that there are two different ways to represent the relevant sets of codepoints.

You can represent what is included or you can represent what is excluded and the representation will depend on which set is smaller for the multibyte groups. 

There is also another problem here and that is related to the individual graphical lexemes which are represented by multiple multibyte codepoints. As far as I understand from the unicode documents, a particular graphical lexeme can be represented by different orderings of the underlying multibyte unicode codepoints. If I understand what the standard is indicating, one lexeme has to treat those different orderings of codepoints as equivalent to each other. This raises the question of what this means for lexicographic equality. 

When I wrote the UTF8 classes, I deliberately chose not to deal with this area and chose that different orderings of the underlying codepoints forced non-equivalence to occur.

In addition, I am currently investigating what kinds of fonts need to be installed to allow the proper graphical display of codepoints within the unicon graphics environment. This is turning into a somewhat hairy problem and at present I haven't got a detailed handle on how to solve this.  

Now the following comment is a personal opinion and comes from someone who is by no means proficient in the c programming language nor in the various vagaries involved in writing efficient c language code. Keeping that in mind, my comment below may be and probably is less than useful.

I have, in the past, looked at various parts of the underlying c language implementation and have noted that there are many places in the code where an appropriate named macro could be used instead of the #ifdef's currently used. I have also noted that there are quite a number of code fragments that are similar in structure but differ in order even though the final result is the same in each case.  I only mention this becasue of your comment above regarding the pre-increment usage of the ++ operator.

I do understand just how difficult it can be to clean up code especially when that code base has been in existence for a long period of time and has had a lot of different programmers involved in its development.

cjeffery:
Bruce, thanks for your comments. At one point I did think about Unicode csets and at that time I did believe in the notion of positive and negative csets. But I don't think anyone has proposed a concrete cset representation for this PR yet.

Jafaral:
Thanks Clint. Replies inline. The corresponding changes are on the unicode branch (UTR 25, `--enable-unicode`, `#ifdef UniconUnicode`, benchmarks, concat, comments).

> The claim of this highly desirable feature (Unicode support) having
> minimal added performance cost are unsubstantiated. [...] Minimally, the
> Unicon benchmarks should be run with and without Unicode support and
> compared. Other evidence giving the user an idea of the cost(s) of the
> feature (primarily time, it would seem) would be welcome.

Done. Numbers are in UTR 25 §10.

`tests/bench` on Linux (Ryzen 9 9950X, gcc 13.3.0): master vs Unicode compiled in. Sequential programs are a wash, mixed ±1–3%. Full-suite wall ~1m23s either way. Concurrent numbers on that host are too noisy to treat as a Unicode signal.

ASCII string ops are a better test of the on-build tax, because they never walk UTF-8. `tests/bench/strascii/` times one untagged operation per program, Unicode off vs on, CPU (`&time`) and wall (`gettimeofday()`), mean of three runs (macOS arm64, clang 21.0.0). Most are 0–1%. `a||b` is +3%; `s[i:j]` is +2%. That is masked `StrLen` plus failed `IsUniQual` / a concat helper that returns immediately — not a UTF-8 scan. Growing `s||:=` is 0%; the append dwarfs two bit tests.

> It is not acceptable to merge this feature without proper #ifdef's
> around everything. If your ifdef is not defined, the original code
> should be present and provided unmolested. If that is the case, all
> your new stuff should just not be compiled in, and not be imposing the
> many additional runtime checks currently in the PR, Unicode or no.

Agreed, and that is now the case. `UniconUnicode` is off unless `./configure --enable-unicode`. Without it, `StrLen`/`SetStrLen` are the historic macros (`StrLen` is an lvalue, no mask). `IsUniQual` is 0. RTT strips the `#ifdef UniconUnicode` blocks, so the Unicode helpers and branches are not in the generated C.

> In utr25.rst, the generous co-author credits should be checked for
> co-author permission. For example, I will claim that I was not
> involved and am not a co-author, unless you can convince me otherwise.
> [...]

Dropped. You, Bruce, and Don are off the list. Author is only me, let me know if you think you should be added.

> Although not as big an issue as the proper #ifdef's being absent,
> turning on Unicode by default raises the bar [...] Calling it
> "optional" when it is on by default is disingenuous. Instructions for
> how to turn it off in the configure step would be helpful [...] But
> perhaps it would be best to start with a period where it is optional
> and the user has to explicitly ask for it.

Off unless the user asks. `./configure --enable-unicode` (64-bit). Documented in UTR 25 §1, `configure --help`, and README next to `--enable-ovld`.

> The discussion of whether to cache a "codepoint count" or a "scan
> position" is interesting, but confusing. A data descriptor encoding a
> property of the data (codepoint count) seems non-dangerous, but one
> encoding a property of a particular computation ("scan position")
> sounds like trouble. That is not just data any more.

We agree. The descriptor caches a codepoint count (a property of the bytes) or leaves it uncached. Scan position is not stored on the string. The index/cache discussion in §9 is future work for a different representation, not a field on the qualifier.

> There is an interesting question about when to cache codepoint counts,
> as described in the discussion of concatenation. [...] If a
> concatenation of a large string with counted codepoints concatenated
> onto a small string without counted codepoints results in discarding a
> large computed result, it might seem cleverer to incrementally update
> the existing codepoint count at concatenation time. The threshold
> where it becomes logical to discard known codepoint count information
> rather than computing an incremental update is a fine question.

That case is handled now. If one side has a known count and the other is tagged-uncached and `StrLen <= 4096`, we walk the short side and keep the sum. Above that, we still drop rather than scan a large unknown. Compile-time cached counts on literals (not only the tag) is still future work; it is listed in UTR §9.

> The discussion of read() and reads() and similar input initially read
> as being "raw only", a puzzling choice. The discussion of "i" mode
> later clarified things, and maybe should come earlier. But "i" is a
> curious mode letter to use. If it stands for something, what is it?
> If it doesn't stand for anything mnemonic, we should consider
> alternatives. It is too bad mode letters seem to be case-insensitive,
> or we could leave "u" as untranslated but take "U" for "Unicode". Oh
> well.

UTR §1 now says up front: reads are untagged unless the program asks (`unicode(s)` or `open(..., "i")`). `i` is international (input as text). `u` is already untranslated. Happy to take another unused letter if there is a better one.

> Discussion at second paragraph of 7.0 of the report, right before 7.1,
> is unclear. Notion that the way to drop a tag is to write the string
> and read it back in again seems silly. Discussion in 7.2 is unclear.

Agreed on the write/read recipe — that paragraph is gone. There is no public inverse yet; internally it is a descriptor copy plus `SetStrLen`. `string()` stays identity, including the tag. §7.2 now states the two questions directly: content (does the payload contain non-ASCII UTF-8?) vs tag (is this descriptor in the text view?). Those can differ for the same bytes.

> Internally, strings get used a lot for more than just * [] and
> move/tab/pos. Is the behavior checked everywhere that the RTL code
> says is:string and so forth? Apparently in many locations, the
> unicode string is still a valid icon string and so you can get away
> with not customizing code, but there are liable to be surprises.

A tagged string is still a string: same blockless qualifier, same bytes. `is:string` does not care about the tag. `StrLen` is the byte length (masked when the feature is on). Hashing and table/set keys compare bytes and ignore the tag. User-visible positions (`*`, `[]`, `!`, `?`, `||`, `reverse`, substring assign, `move`/`tab`/`pos`, `find`/`match`/`many`) were updated. Remaining string-analysis functions that take a cset are documented as still byte-oriented (`any`/`upto`/`bal`/`map`/…). I have not claimed a line-by-line audit of every `is:string` in the RTL; surprises there would be bugs, and I want those reports.

> Similarly, do the various StrLoc(x)++ operations in the runtime need
> to be modified to advance one whole codepoint if UniQual, with
> corresponding StrLen() modifications, or can we prove they are never
> UniQual? See fwindow.r

They must stay byte advances. Those sites are not user-visible indices. They strip an ASCII quote, skip to the next newline in a console buffer, or parse an ASCII attribute string (`fwindow.r` walks `"width,data"` with `s++`). `"` and `\n` are never UTF-8 continuation bytes, so skipping them with `StrLoc++` is correct even on a tagged value. Advancing a whole codepoint there would be wrong.

> The work is incomplete vis a vis csets and the string analysis
> functions that use them.

Yes. Documented in UTR §9. A 256-bit cset cannot represent codepoints above 255. `many` already walks tagged subjects by codepoint but has nothing to match a non-ASCII codepoint against. I am not taking that on in this PR. Bruce's include-vs-exclude / grapheme notes belong with that later design, not here.

> You've changed StrLen(d) to use a bitwise AND, and this change appears
> to apply whether Unicode is compiled in or not. This is not OK.

Fixed with the `#ifdef`. Off-build `StrLen` is unmolested.

> For StrLen(d) and SetStrLen(), we need to understand if
> bitwise-and'ing on 64-bit operations is a faster option than doing
> direct/native 32-bit operations. [...] in cnv.r you turn a ++StrLen(d)
> into a SetStrLen() and I am wondering if a casted 32-bit ++ would
> still be faster.

Off-build there is no AND and `++StrLen` can stay an lvalue. On-build the mask is required because the dword holds the tag and cached count; a 32-bit `++` through a cast would not be safe. The ASCII microbenchmarks are the cost of that on-build path: 0–3% on these operations. I have not compared a hypothetical 32-bit increment against the mask on a Unicode build, because that increment cannot preserve the high bits.

> One exception to this is: it would be possible to define a SetStrLen()
> and introduce its use everywhere needed, separate and prior to the
> main PR. The MakeStr() macro could similarly be introduced as part of
> the separate/prior PR. [...]

`SetStrLen` and the `MakeStr` sweep are already their own commits in this stack. I can split them into a prior PR if that is the merge path you want. Enforcing the discipline on later patches is a real issue either way.

> "Unicon Phase 0" language in comments in various files is confusing.
> Do you mean Unicode Phase 0? In fstranl.r it is even worse, just
> saying "Unicon:" in places, not even "Unicon Phase 0".

Renamed to `Unicode:`. Same for the `Unicon:` stubs in `fstranl.r` / `fstr.r` / `interp.r`.

> Comments are needed where you are tending new variables in fsys.r --
> why is the new tended variable needed in those places? Is it that
> RTT's return string(x,y) doesn't know to do the correct thing yet?
> If so why not? And why is the new variable tended in these various
> spots? Are you introducing additional allocations somehow?

Comments are in. `return string(p, n)` builds the qualifier at the return site, so there is no value left to tag. Those paths `MakeStr` into a named `uq_result`, call `UqMaybeTagRead`, then return it. No extra allocation: `MakeStr` points at bytes already in string space. It is tended because it is a qualifier into string space (and in the pty `reads()` path a local `char *s` already shadows the outer descriptor).

Jafaral:
@cjeffery @brucerennie, I intentionally left csets out in this round. I do have old design notes from the 3 of us on this topic, and initially I wanted to do it as part of this PR. However, I quickly discovered that most of the work needed to support Unicode for csets is independent of string supportat almost all runtime/language levels. So It is easier to leave that alone until we are happy with strings!

Jafaral:
Section 10 contains performance numbers: https://jafaral.github.io/unicon/doc/utr/html/utr25.html

PR #654, opened 2026-08-17
ide: Stop utags from looping on link (

### LINKED ISSUE #651: Memory problem when opening the file parsedprogram.icn in ui
Testing the latest changes for the modern pull request, I used ui to open all of the ".icn" files in uni/parser. The program could open all but one of the source files located in this directory with a total memory allocation of about 93 Mbytes.

However, when opening the source file parsedprogram.icn, even alone, the memory usage went for 53 Mbytes to 1.7 Gbytes. I will undertake an removal of lines from this file to see what might be causing this explosion.

Jafaral:
getword() treated '(' as a word start but never consumed it, so outline scanning allocated tags forever on lines such as "link" : add_link(decl).

fixes: #651

### DIFF
--- a/uni/ide/utags.icn
+++ b/uni/ide/utags.icn
@@ -28,14 +28,19 @@ procedure utags(contents, parentfile)
       line[find("//",line):0] := ""
 
       line ? {
-         gword := getword3(line)
 ### Unicon Tags ###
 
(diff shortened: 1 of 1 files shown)


brucerennie:
Jafar,

As far as I can tell, this fixes the particular problem in the reported bug. Thank you.

PR #658, opened 2026-08-18
Docupdates - correcting a number of spelling mistakes and format fixes in Unicon book

brucerennie:
A number of spelling mistakes have been corrected using the US language as the reference. 

A number of format corrections to bring a variety of terms into the same format.


cjeffery:
Thank you for these fixes and improvements. All looks good to me if greptile issues have been corrected. @greptile review

PR #668, opened 2026-09-15
Ulsp - standardise table initialisation and standard Unicon function names

brucerennie:
The table initialisation  has been standardised across all the files contained in the ulsp package,

changed the various uses of table(key, value, ...) to the format [key : value; ...]

updated all references to the standard unicon functions from func(...) to ::func(...) 

This pull request is a cosmetic code update and is optional as to whether or not it is actually installed into the master branch of the unicon project.  I found it was much easier to read the code for the ulsp package by doing this. 

Any discussion for this pull request is welcome.


Don-Ward (commented):
The PR is not about changing functionality but instead (as is admitted) more about coding style. Thus its acceptance is more a matter of opinion than one of correctness -- unless it violates our coding standards, of which there are few.

Having said that, I don't find the editted text easier to read than the original version.  I usually treat the `:: `as an explicit warning that the global identifier in question has been redefined locally and the `:: `makes it clear that the global definition is required, rather than the local one. When there is no redefinition the warning is not required and, to me, is just uneccessary noise. Obviously, Bruce doesn't agree with me. 

 If all, or even most, of the other unicon source were coded with `:: ` and the PR was making  this code conform with the rest, I'd be in favour of it. But that isn't the case; if anything the reverse is true.

The changes in table initialisation are probably a good thing (in my opinion) not because they are easier to read -- I find the two versions about the same -- but because it's harder to inadvertantly make a mistake by inserting a spurious comma.

brucerennie:
@Don-Ward 

Thank you for your feedback. 

From my perspective, it is important that I understand what others perceive as being more readily comprehensible within the code base. So when you say [Having said that, I don't find the editted text easier to read than the original version.], this, in effect, clarifies for me what others might or might not find easier to read and comprehend.

In the particular case as you state [I usually treat the :: as an explicit warning that the global identifier in question has been redefined locally and the :: makes it clear that the global definition is required, rather than the local one.], I am in agreement with you. However, I like clarity as to where a reference is pointing to, the current package, an imported package or the global environment.

I have found in my own coding that certain problems have arisen when [importing] other packages. This has caught me out a couple times and as such, I tend now to directly reference the associated package for any class/procedure/method/record even with an import directive, including the global environment.

This particular problem may be very much limited to what I do and not at all apply to anyone else.

I do appreciate that this may be [just uneccessary noise] for those who are long time users of the language. Where I am coming from here is looking at the perspective of new Unicon users. I would very much like to see more people taking up the language as a standard tool. 

If one goes through the IPL and the UNI sets of files, there is a large variety of expression construction. Some of it is quite idiomatic in terms of the basic ideas from icon and Unicon. But there is also plenty of code that is, how shall we put this, written in idiomatic BASIC or FORTRAN. Yes that is hyperbole, but I think you will get what I mean. Even withing the same file or procedure, an idiomatic form may be used in one place and a non-idiomatic form used nearby.

As you say Don, [If all, or even most, of the other Unicon source were coded with :: and the PR was making this code conform with the rest, I'd be in favour of it. But that isn't the case; if anything the reverse is true.] and one of my longer term projects is to attempt to update the entire code base to be more idiomatic and document the various forms that can be used. Again, this PR is more for discussion and feedback than being pushed to get these changes in.

As we have seen in the past, there are various members of the community that have quite strong feelings about coding style as well as other aspects applicable to Unicon and the available facilities. This is actually a very good thing as it allows all of us to consider alternatives to the specific ways we as individuals might write Unicon code.

As for the table initialisation, I personally find the new form much easier to read and comprehend. The table(...) call form is for me somewhat discordant and hides what is going on here. 

So again, Don, thank you for your feedback. It is much appreciated.

Jafaral:
I do like the table format change. I do agree that format is much more readable than using `table()`. I also agree with Don's opinion. `put()` looks much better than `::put()`, especially with functions as basic as `put()`.

brucerennie:
@Jafaral @Don-Ward @cjeffery  How about I close this PR and create a new PR that only has the table() change? If there is agreement on that, I will make the changes this weekend. I wait to here from each of you.

Have a blessed weekend.

By the way there is a small error in one of the files when looking for source files with the .icn suffix. The code assumes that "." only occurs once in the file name and doesn't actually look for ".icn" as the suffix. I will create a PR for this next week.

Don-Ward:
The proposal to just implement the table changes works for me.  But, might it be worth waiting for the opinions of the other reviewers?  It could be that Jafar and I are a vocal minority and the consensus goes the other way.

brucerennie:
Until there is agreement one way or the other, I won't be generating the PR. I will however be looking at the other problem I found.

